PR #3, opened 2022-03-02
fix: run go mod tidy and add go.sum

ajcasagrande:
run go mod tidy and add go.sum

Signed-off-by: Anthony Casagrande <anthony.j.casagrande@intel.com>


lenny-goodell (approved):
LGTM

PR #44, opened 2022-06-23
Revert "build: Upgrade sdk and paho.mqtt lib"

cloudxxx8:
Reverts edgexfoundry/device-usb-camera#42

Revert the changes because this will be released with v2.2, and we will update it again after release


cloudxxx8:
close this PR because I can't push the commit to fix DCO in this branch

PR #57, opened 2022-07-22
build: Add rtsp-simple-server LICENSE

### LINKED ISSUE #53: License of rtsp-simple-server missing from Docker image
The docker image does not include the [license](https://github.com/aler9/rtsp-simple-server/blob/main/LICENSE) of rtsp-simple-server. Only the executable and configuration file are packaged:
https://github.com/edgexfoundry/device-usb-camera/blob/d52eed5e0bdf095fe71ab1c2620e174378e09b3d/Dockerfile#L36-L52

# 🐞 Bug Report

Device USB Camera

### Is this a regression?
No

## 🌍  Your Environment
**Deployment Environment:**
Docker

**EdgeX Version [**REQUIRED**]:**
v0.0.1-dev.13

**Anything else relevant?**

FelixTing:
fix: #53 

Signed-off-by: Felix Ting <felix@iotechsys.com>

## PR Checklist

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**If your build fails** due to your commit message not passing the build checks, please review the guidelines here: https://github.com/edgexfoundry/edgex-examples/blob/master/.github/CONTRIBUTING.md.

## What is the current behavior?

The docker image does not include the [license](https://github.com/aler9/rtsp-simple-server/blob/main/LICENSE) of rtsp-simple-server.

## What is the new behavior?
Fetch the LICENSE file from rtsp-simple-server github repo during Docker build.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## New Imports
Are there any new imports or modules? If so, what are they used for and why? N/A

## Specific Instructions
Are there any specific instructions or things that should be known prior to reviewing? N/A

## Other information

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ ARG BASE=golang:1.17-alpine3.15
 FROM ${BASE} AS builder
 
 ARG MAKE="make build"
-ARG ALPINE_PKG_BASE="make git gcc libc-dev zeromq-dev libsodium-dev"
+ARG ALPINE_PKG_BASE="make git gcc libc-dev zeromq-dev libsodium-dev curl"
 ARG ALPINE_PKG_EXTRA="v4l-utils-dev v4l-utils v4l-utils-libs linux-headers"
 
 RUN sed -e 's/dl-cdn[.]alpinelinux.org/dl-4.alpinelinux.org/g' -i~ /etc/apk/repositories
@@ -30,6 +30,8 @@ COPY go.mod vendor* ./
 RUN [ ! -d "vendor" ] && go mod download all || echo "skipping..."
 
 COPY . .
+
+RUN curl -o LICENSE-rtsp-simple-server https://raw.githubusercontent.com/aler9/rtsp-simple-server/main/LICENSE
 
 RUN ${MAKE}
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -31,6 +31,8 @@ RUN [ ! -d "vendor" ] && go mod download all || echo "skipping..."
 
 COPY . .
 
+RUN /device-usb-camera/fetch-license.sh
+
```
farshidtz, line 35: Why not add this to the rtsp stage?

fetch-license.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+curl -o LICENSE-rtsp-simple-server https://raw.githubusercontent.com/aler9/rtsp-simple-server/main/LICENSE
```
farshidtz: I suggest putting the curl command (without the `-o`) in the Dockerfile itself, in rtsp stage, and renaming it in the last stage when copying to the final stage.

It's a bit of an overkill to have a script with this command.

farshidtz: If you insist on keeping it in a script, then the script should be under `bin` and have a more specific name, e.g. `bin/fetch-rtsp-license.sh`.

farshidtz (changes requested):
Thanks for the fix. Please see comments.

fetch-license.sh
FelixTing: Thanks @farshidtz
Removed the script. However, it seems that we can only get the license file at the builder stage. Because rtsp-simple-server image is based on `scrach`: https://github.com/aler9/rtsp-simple-server/blob/f7c08f577af3db5032393b9e840dc43c0b385e0c/Makefile#L230

farshidtz: Thanks. That's okay then. 
In the meantime, I'll check if I can upstream the license addition.

farshidtz: Thanks. That's okay then. 
In the meantime, I'll check if I can upstream the license addition.

farshidtz (approved):

PR #63, opened 2022-07-26
fix: Document fix to build errors for missing prerequisite

### LINKED ISSUE #61: Document need to run "sudo dpkg -i linux-libc-dev_5.10.0-14.15_amd64.deb" in README
# 📚 Docs or Wiki Bug Report

### Description [**REQUIRED**]

User may need to run `Document need to run `sudo dpkg -i linux-libc-dev_5.10.0-14.15_amd64.deb` in order to build the device service.

This needs to be documented in the README.

See this PR comment https://github.com/edgexfoundry/device-usb-camera/issues/10#issuecomment-1068739016

lenny-goodell:
Signed-off-by: Leonard Goodell <leonard.goodell@intel.com>

## PR Checklist

- [ ] I have added unit tests for the new feature or bug fix (if not, why?) **N/A for docs PR**
- [ ] I have fully tested (add details below) this the new feature or bug fix (if not, why?) **N/A for docs PR**
- [ ] I have opened a PR for the related docs change (if not, why?) **N/A for docs PR**
  <link to docs PR>

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,9 +39,13 @@ Note if your build fails with errors like:
 .../go4vl@v0.0.2/v4l2/capability.go:46:33: could not determine kind of name for C.V4L2_CAP_META_OUTPUT
 ```
 
-Try running the following in your Linux terminal
+You are missing the appropriate kernal headers needed by the `github.com/vladimirvivien/go4vl` module
+One possible solution is to manually download and install a more recent version of the libc-dev for your OS.
+
+In the case of Ubuntu 20.04, one is not available in the normal repositories, so you can get it via these steps:
 
 ```
+wget https://launchpad.net/~canonical-kernel-team/+archive/ubuntu/bootstrap/+build/20950478/+files/linux-libc-dev_5.10.0-14.15_amd64.deb
 sudo dpkg -i linux-libc-dev_5.10.0-14.15_amd64.deb
 ```
 
--- a/README.md
+++ b/README.md
@@ -32,7 +32,21 @@ The following devices have been tested with EdgeX:
 make build
 ```
 
+Note if your build fails with errors like:
+
+```
+.../go4vl@v0.0.2/v4l2/capability.go:48:33: could not determine kind of name for C.V4L2_CAP_IO_MC
+.../go4vl@v0.0.2/v4l2/capability.go:46:33: could not determine kind of name for C.V4L2_CAP_META_OUTPUT
+```
+
+Try running the following in your Linux terminal
+
+```
+sudo dpkg -i linux-libc-dev_5.10.0-14.15_amd64.deb
+```
+
 ### Build docker image
+
 ```shell
 make docker
 ```


README.md
```diff
@@ -32,7 +32,21 @@ The following devices have been tested with EdgeX:
...
+
+```
+.../go4vl@v0.0.2/v4l2/capability.go:48:33: could not determine kind of name for C.V4L2_CAP_IO_MC
+.../go4vl@v0.0.2/v4l2/capability.go:46:33: could not determine kind of name for C.V4L2_CAP_META_OUTPUT
+```
+
+Try running the following in your Linux terminal
+
+```
+sudo dpkg -i linux-libc-dev_5.10.0-14.15_amd64.deb
```
ajcasagrande, line 49: This command says to install a debian package from the filesystem. Where do they get this file from? How does it affect your system? What if you aren't using Debian/Ubuntu? Lots of holes with this. Also, can this build process be dockerized so the user doesn't need to muck with their system?

ajcasagrande (changes requested):

README.md
lenny-goodell: This is just for local build section. Docker builds fine.

This was the fix from here: https://github.com/vladimirvivien/go4vl/issues/17
@surajitx-pal confirmed it worked on his system.

What addition context/restrictions should be added.

ajcasagrande: A few things:
- The scope of systems should be narrower. Do we recommend it for Ubuntu 20.04 only users? Users of kernel version X.XX-X.XX?
- Missing download link. Command does nothing without the actual file to install
- What are the ramifications of installing these headers?
- Are the missing definitions truly just constants or will they cause parts of the code to work in unexpected ways?

lenny-goodell: So better to update README to recommend `Ubuntu 20.04` and then in build note instruct them to use `Ubuntu 20.04` to resolve the errors? i.e. they are not on `kernel v5.10 or higher` if they are getting these errors...

ajcasagrande: @lenny-intel I've done some digging on this and it turned out to be way more convoluted than I had originally assumed. It turns out that in the specific case of these, the kernel version matters less than the version of the kernel headers in use.

I think `Ubuntu 22.04` is actually immune from this issue, as it is more up-to-date. On my VM with `Ubuntu 20.04`, I had kernel version `5.13.x`, but the `libc-dev` package only exists as `5.4.x` as apparently that is how these things are handled from what I read? Regardless, the build still fails due to the old headers.

(Side note, but my dev machine uses Manjaro with a 5.17.x kernel with no issue since it is up-to-date as well).

Back to the VM. Sure, updating the headers worked, using the trick to download and install that deb file, however it would make most sense to use a version similar or matching the installed kernel (eg 5.13.x).

Going further, with `libc-dev` being held back to `5.4.x`, there is the option to install `sudo apt-get install linux-headers-$(uname r)`, however that installs them in a place not automatically found by `go4vl`.

To make matters even more strange, According to [docs.kernel.org](https://docs.kernel.org/userspace-api/media/v4l/func-ioctl.html#description)
> Macros and defines specifying V4L2 ioctl requests are located in the videodev2.h header file. Applications should use their own copy, not include the version in the kernel sources on the system they compile on.

So maybe go4vl is supposed to come with its own copy of that header file?

ajcasagrande: Related PR has been opened in `go4vl` library:
https://github.com/vladimirvivien/go4vl/pull/20

README.md
```diff
@@ -32,7 +32,21 @@ The following devices have been tested with EdgeX:
...
 ```
 
+Note if your build fails with errors like:
+
+```
+.../go4vl@v0.0.2/v4l2/capability.go:48:33: could not determine kind of name for C.V4L2_CAP_IO_MC
+.../go4vl@v0.0.2/v4l2/capability.go:46:33: could not determine kind of name for C.V4L2_CAP_META_OUTPUT
+```
+
+Try running the following in your Linux terminal
```
ajcasagrande: One possible solution is to manually download and install a more recent version of the `libc-dev` package for your OS.

In the case of `Ubuntu 20.04`, one is not available in the normal repositories, so you can get it via these steps:

```shell
wget https://launchpad.net/~canonical-kernel-team/+archive/ubuntu/bootstrap/+build/20950478/+files/linux-libc-dev_5.10.0-14.15_amd64.deb
sudo dpkg -i linux-libc-dev_5.10.0-14.15_amd64.deb
```

lenny-goodell: @ajcasagrande , THX!

ajcasagrande (approved):

PR #194, opened 2023-03-21
fix(snap): Refactor to avoid conflicts with readonly config provider directory

MonicaisHer:
This PR makes changes to incorporate the following fixes:
- https://github.com/canonical/edgex-snap-hooks/pull/83
- https://github.com/canonical/edgex-snap-hooks/pull/82

## PR Checklist

  <link to docs PR>

## Testing Instructions

For testing instructions, please refer to https://github.com/edgexfoundry/device-mqtt-go/pull/535

### DIFF
--- a/snap/local/helper-go/install.go
+++ b/snap/local/helper-go/install.go
@@ -20,6 +20,7 @@ import (
 	hooks "github.com/canonical/edgex-snap-hooks/v3"
 	"github.com/canonical/edgex-snap-hooks/v3/env"
 	"github.com/canonical/edgex-snap-hooks/v3/log"
+	"github.com/canonical/edgex-snap-hooks/v3/snapctl"
 )
 
 func installUSBCameraConfig() error {
@@ -37,8 +38,15 @@ func installRTSPConfig() error {
 func install() {
 	log.SetComponentName("install")
 
-	if err := installUSBCameraConfig(); err != nil {
-		log.Fatalf("Error installing Device USB Camera files: %s", err)
+	// Install default config files for Device USB Camera only if no config provider is connected
+	isConnected, err := snapctl.IsConnected(usbCameraApp + "-config").Run()
+	if err != nil {
+		log.Fatalf("Error checking interface connection: %s", err)
+	}
+	if !isConnected {
+		if err := installUSBCameraConfig(); err != nil {
+			log.Fatalf("Error installing Device USB Camera files: %s", err)
+		}
 	}
 
 	if err := installRTSPConfig(); err != nil {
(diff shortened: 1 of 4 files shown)


snap/local/helper-go/install.go
```diff
@@ -37,11 +38,25 @@ func installRTSPConfig() error {
...
+	if !isConnected {
+		if err := installUSBCameraConfig(); err != nil {
+			log.Fatalf("Error installing Device USB Camera files: %s", err)
+		}
 	}
 
-	if err := installRTSPConfig(); err != nil {
-		log.Fatalf("Error installing RTSP Simple Server config file: %s", err)
+	// Install default config files for RTSP server app only if no config provider is connected
+	isConnected, err = snapctl.IsConnected(rtspServerApp + "-config").Run()
```
farshidtz: There is no such plug in the snap. I'm surprised that this isn't causing an installation error, but maybe it is actually skipping the copying of files.

farshidtz (changes requested):

snap/local/helper-go/install.go
MonicaisHer: I believe this is because there is no corresponding plug available, causing the `IsConnected` to return false. As a result, the install hook will proceed to copy files.

MonicaisHer: If a non-existent plug is passed to `snapctl is-connected`, the command will return exit status code 1:
```
$ snapctl is-connected non-exist-plug
error: error running snapctl: snap "edgexfoundry" has no plug or slot named "non-exist-plug"
$ echo $?
1
```

farshidtz: I expect that would fall into [this](https://github.com/canonical/edgex-snap-hooks/blob/544c6823a1b007cf18e22cfa85d1ba12b0ffa2ba/snapctl/is-connected.go#L82) and return the error, but maybe there is a bug.

farshidtz: Regardless, this PR should not check if a non-existing plug is connected.

MonicaisHer: ok, I reverted install hook related changes

farshidtz: I think you misunderstood my comments. Your code was checking two interfaces, one called `rtsp-simple-server-config` which doesn't exist, another called `device-usb-camera-config` which does exist and was correct. Please see the following part of the snapcraft.yaml file to know what content interface is defined for this snap, and need the fix: https://github.com/edgexfoundry/device-usb-camera/blob/96327e049a2d26e356251a4ee56d80af5c884d85/snap/snapcraft.yaml#L21

farshidtz (approved):

PR #229, opened 2023-04-28
Rtsp cred

brianointel:

## PR Checklist

- [ ] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [ ] I am not introducing a new dependency (add notes below if you are)
  <link to docs PR>


internal/driver/driver.go
```diff
@@ -62,6 +63,16 @@ type Driver struct {
 	addedWatchers bool
 	watchersMu    sync.Mutex
 }
+type RtspRequest struct {
```
ajcasagrande: i would move this into a separate file maybe `types.go`

internal/driver/driver.go
```diff
@@ -148,8 +159,71 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
 	// Make sure the paths of existing devices are up-to-date.
 	go d.RefreshExistingDevicePaths()
 
+	http.HandleFunc("/rtspauth", RtspCredentialsHandler)
```
ajcasagrande: i think this call can go inside the `d.StartRtspCredentialServer()` function

internal/driver/driver.go
```diff
@@ -148,8 +159,71 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
+	go d.StartRtspCredentialServer()
 	return nil
 }
+func (d *Driver) StartRtspCredentialServer() {
+	d.lc.Infof("starting rtsp server\n")
+	err := http.ListenAndServe(":8000", nil) //TODO: get port from config
```
ajcasagrande: we need to either set the host part as `localhost` or have a config option for that as well.

internal/driver/driver.go
```diff
@@ -148,8 +159,71 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
 	// Make sure the paths of existing devices are up-to-date.
 	go d.RefreshExistingDevicePaths()
 
+	http.HandleFunc("/rtspauth", RtspCredentialsHandler)
+	go d.StartRtspCredentialServer()
 	return nil
 }
+func (d *Driver) StartRtspCredentialServer() {
+	d.lc.Infof("starting rtsp server\n")
```
ajcasagrande: I think it might be better to do something like this, so it doesn't use the default http objects (in case the sdk uses it for something).

```go
    rtspAuthServer := http.NewServeMux()
    rtspAuthServer.HandleFunc("/rtspauth", d.rtspCredentialsHandler)
    http.ListenAndServe("localhost:8080", rtspAuthServer)
```

internal/driver/driver.go
```diff
@@ -148,8 +159,71 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
+	err := http.ListenAndServe(":8000", nil) //TODO: get port from config
+	if err == http.ErrServerClosed {
+		d.lc.Errorf("server closed\n")
+	} else if err != nil {
+		d.lc.Errorf("error starting server: %s\n", err)
+	}
+}
+func RtspCredentialsHandler(w http.ResponseWriter, r *http.Request) {
```
ajcasagrande: make this a receiver function of the driver class to allow it to use `d.lc` and such.

internal/driver/driver.go
```diff
@@ -148,8 +159,71 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
+	if err != nil {
+		driver.lc.Warnf("could not unmarshal read body: %s\n", err)
+		w.WriteHeader(http.StatusUnauthorized)
+		return
+	}
+	fmt.Printf("user read body: %s\n", rtspRequest.User)
+	if rtspRequest.Action == "publish" {
+		driver.lc.Infof("allowing publish at path %s", rtspRequest.Path)
+		w.WriteHeader(http.StatusOK)
+	}
```
ajcasagrande: is this skipping the auth check for publish? I think we can still do the auth check, and just have the publishing code provide the username and password.

internal/driver/driver.go
```diff
@@ -148,8 +159,71 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
+	credential, edgexErr := driver.tryGetCredentials("rtsp_" + rtspRequest.User)
```
ajcasagrande: i think the secret name should probably be a hard-coded constant. If we wanted to support more than one credential, it would probably be on a per-camera basis, so like based on the `rtspRequest.Path`. For now we can stick with one, and we can add the more advanced option if we see fit.

internal/driver/driver.go
```diff
@@ -148,8 +159,71 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
+	if edgexErr != nil {
+		driver.lc.Warnf("failed to get credentials for at path %s", rtspRequest.Path)
```
ajcasagrande: replace all usages of `driver.lc` with `d.lc` once it is a receiver function`

internal/driver/driver.go
```diff
@@ -148,8 +159,71 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
+		w.WriteHeader(http.StatusUnauthorized)
+		return
+	}
+	// TODO: move to seperate function?
+	if credential.Password == rtspRequest.Password {
+		fmt.Printf("passwords match\n")
```
ajcasagrande: change all prints to debug logs using `d.lc`

internal/driver/driver.go
```diff
@@ -148,8 +159,71 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
+		w.WriteHeader(http.StatusOK)
+	}
+	credential, edgexErr := driver.tryGetCredentials("rtsp_" + rtspRequest.User)
+	if edgexErr != nil {
+		driver.lc.Warnf("failed to get credentials for at path %s", rtspRequest.Path)
+		w.WriteHeader(http.StatusUnauthorized)
+		return
+	}
+	// TODO: move to seperate function?
+	if credential.Password == rtspRequest.Password {
```
ajcasagrande: we will need to check the username matches as well

internal/driver/driver.go
ajcasagrande: i suggest maybe we go with a name like `rtspauth_default` or just `rtspauth` as the hard coded one, and then in the future we can expand it to support `rtspauth_<pathname>`, etc.

internal/driver/driver.go
```diff
@@ -148,8 +159,71 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
+		driver.lc.Infof("allowing publish at path %s", rtspRequest.Path)
+		w.WriteHeader(http.StatusOK)
+	}
+	credential, edgexErr := driver.tryGetCredentials("rtsp_" + rtspRequest.User)
+	if edgexErr != nil {
+		driver.lc.Warnf("failed to get credentials for at path %s", rtspRequest.Path)
+		w.WriteHeader(http.StatusUnauthorized)
+		return
+	}
+	// TODO: move to seperate function?
```
ajcasagrande: same function is ok with me

ajcasagrande (changes requested):

internal/driver/types.go
```diff
@@ -0,0 +1,18 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+//
+// Copyright (C) 2022 Intel Corporation
```
ajcasagrande: 2023

internal/driver/driver.go
```diff
@@ -562,6 +617,15 @@ func (d *Driver) newDevice(name string, protocols map[string]models.ProtocolProp
 		Scheme: RtspUriScheme,
 		Host:   fmt.Sprintf("%s:%s", d.rtspHostName, d.rtspTcpPort),
 	}
+	credential, edgexErr := d.tryGetCredentials(rtspAuthKey)
+	if edgexErr != nil {
+		d.lc.Warnf("failed to get credentials for at path %s", rtspAuthKey)
+	}
+
+	if credential.Username != "" && credential.Password != "" {
+		rtspUri.User = url.UserPassword(credential.Username, credential.Password)
```
ajcasagrande, line 626: i do not think we want to cache this here

internal/driver/types.go
```diff
@@ -0,0 +1,18 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package driver
+
+type RtspRequest struct {
```
ajcasagrande: Go uses uppercase for acronyms, also should add Auth in there:

`RTSPAuthRequest` or `RTSPAuthenticationRequest`

internal/driver/driver.go
```diff
@@ -562,6 +617,15 @@ func (d *Driver) newDevice(name string, protocols map[string]models.ProtocolProp
 		Scheme: RtspUriScheme,
 		Host:   fmt.Sprintf("%s:%s", d.rtspHostName, d.rtspTcpPort),
 	}
+	credential, edgexErr := d.tryGetCredentials(rtspAuthKey)
+	if edgexErr != nil {
+		d.lc.Warnf("failed to get credentials for at path %s", rtspAuthKey)
```
ajcasagrande: reword. something like so:
```suggestion
		d.lc.Warnf("Failed to get credentials for rtsp authentication from secretName %s", rtspAuthKey)
```

internal/driver/credentials.go
```diff
@@ -0,0 +1,40 @@
...
+
+// Credentials encapsulates Username and Password pair.
+type Credentials struct {
+	Username string
+	Password string
+}
+
+const (
+	UsernameKey = "username"
+	PasswordKey = "password"
```
ajcasagrande, line 21: these are already defined in go-mod-bootstrap `secret` package

internal/driver/driver.go
```diff
@@ -147,9 +151,54 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
 	go d.RefreshExistingDevicePaths()
-
+	go d.StartRtspCredentialServer()
 	return nil
 }
+func (d *Driver) StartRtspCredentialServer() {
+	d.lc.Infof("starting rtsp server\n")
+	rtspAuthServer := http.NewServeMux()
+	rtspAuthServer.HandleFunc("/rtspauth", d.RtspCredentialsHandler)
+	http.ListenAndServe("localhost:8000", rtspAuthServer)
```
ajcasagrande: do we need to shut down the listener on driver.Stop?

internal/driver/driver.go
```diff
@@ -147,9 +151,54 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
+}
+func (d *Driver) RtspCredentialsHandler(w http.ResponseWriter, r *http.Request) {
```
ajcasagrande: RTSP shouldbe either all uppercase or all lowercase `RTSP` or `rtsp`, but not `Rtsp`. See: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

internal/driver/driver.go
```diff
@@ -147,9 +151,54 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
+
+	credential, edgexErr := d.tryGetCredentials(rtspAuthKey)
+	if edgexErr != nil {
+		d.lc.Warnf("failed to get credentials for at path %s", rtspRequest.Path)
+		w.WriteHeader(http.StatusUnauthorized)
+		return
+	}
+
+	if credential.Username == "" && credential.Password == "" {
+		d.lc.Debug("rtsp password disabled\n")
```
ajcasagrande: remove the `\n` from these logs.

internal/driver/driver.go
```diff
@@ -147,9 +151,54 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
+	if err != nil {
+		d.lc.Warnf("could not read body: %s\n", err)
+		w.WriteHeader(http.StatusUnauthorized)
+		return
+	}
+
+	var rtspRequest RtspRequest
+	err = json.Unmarshal(body, &rtspRequest)
+	if err != nil {
+		d.lc.Warnf("could not unmarshal read body: %s\n", err)
```
ajcasagrande: Errorf

internal/driver/driver.go
```diff
@@ -147,9 +151,54 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
+	d.lc.Infof("starting rtsp server\n")
+	rtspAuthServer := http.NewServeMux()
+	rtspAuthServer.HandleFunc("/rtspauth", d.RtspCredentialsHandler)
+	http.ListenAndServe("localhost:8000", rtspAuthServer)
+}
+func (d *Driver) RtspCredentialsHandler(w http.ResponseWriter, r *http.Request) {
+
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		d.lc.Warnf("could not read body: %s\n", err)
```
ajcasagrande: Errorf

internal/driver/driver.go
```diff
@@ -147,9 +151,54 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
+	err = json.Unmarshal(body, &rtspRequest)
+	if err != nil {
+		d.lc.Warnf("could not unmarshal read body: %s\n", err)
+		w.WriteHeader(http.StatusUnauthorized)
+		return
+	}
+
+	credential, edgexErr := d.tryGetCredentials(rtspAuthKey)
+	if edgexErr != nil {
+		d.lc.Warnf("failed to get credentials for at path %s", rtspRequest.Path)
```
ajcasagrande: reword. also, we should probably add a message hinting the user that they might not have set the credentials yet:
```suggestion
		d.lc.Warnf("Failed to retrieve credentials for rtsp authentication from the secret store. Have you stored credentials yet for secretName %s?", rtspAuthKey)
```

internal/driver/driver.go
```diff
@@ -147,9 +151,54 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
+		d.lc.Warnf("could not read body: %s\n", err)
+		w.WriteHeader(http.StatusUnauthorized)
+		return
+	}
+
+	var rtspRequest RtspRequest
+	err = json.Unmarshal(body, &rtspRequest)
+	if err != nil {
+		d.lc.Warnf("could not unmarshal read body: %s\n", err)
+		w.WriteHeader(http.StatusUnauthorized)
```
ajcasagrande: this should probably be a 400 BadRequest.

internal/driver/driver.go
```diff
@@ -147,9 +151,54 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
+	rtspAuthServer := http.NewServeMux()
+	rtspAuthServer.HandleFunc("/rtspauth", d.RtspCredentialsHandler)
+	http.ListenAndServe("localhost:8000", rtspAuthServer)
+}
+func (d *Driver) RtspCredentialsHandler(w http.ResponseWriter, r *http.Request) {
+
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		d.lc.Warnf("could not read body: %s\n", err)
+		w.WriteHeader(http.StatusUnauthorized)
```
ajcasagrande: should probably be a 400 Bad Request or 500 Internal Error.

internal/driver/driver.go
ajcasagrande: didn't you already have the option to make the port configurable? I think we need that.

internal/driver/driver.go
```diff
@@ -47,6 +47,10 @@ const (
 	provisionWatcherRetrySleep = 200 * time.Millisecond
 )
 
+const (
+	rtspAuthKey string = "rtspauth"
```
ajcasagrande, line 52: comment on what this is used for

cmd/res/configuration.toml
```diff
@@ -10,6 +10,11 @@ LogLevel = "INFO"
       [Writable.InsecureSecrets.DB.Secrets]
       username = ""
       password = ""
+    [Writable.InsecureSecrets.rtspauth]
+    path = "rtspauth"
+      [Writable.InsecureSecrets.rtspauth.Secrets]
+      username = "rtspauth"
+      password = "Passwrd123"
```
ajcasagrande: do not commit a default. leave the values as blank

cmd/res/configuration.toml
```diff
@@ -127,10 +132,10 @@ Labels = []
 UseMessageBus = true
   [Device.Discovery]
   Enabled = true
-  Interval = "1h"
+  Interval = "30s"
 
 [Driver]
-RtspServerHostName = "localhost"
+RtspServerHostName = "192.168.10.4"
```
ajcasagrande: undo your local dev changes. Note, for development it is often easier to use env overrides so that you do not change the file

internal/driver/driver.go
```diff
@@ -233,7 +282,13 @@ func (d *Driver) HandleReadCommands(deviceName string, protocols map[string]mode
 			}
 			cv, err = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeObject, data)
 		case VideoStreamUri:
-			cv, err = sdkModels.NewCommandValue(req.DeviceResourceName, req.Type, device.rtspUri)
+			rtspUri := &url.URL{
```
ajcasagrande: i think we should revert this back to how it is, and when we go to start streaming do the user/pass lookup there instead.

ajcasagrande (changes requested):

internal/driver/constants.go
```diff
@@ -20,6 +20,8 @@ const (
 	DefaultRtspServerHostName  = "localhost"
 	RtspTcpPort                = "RtspTcpPort"
 	DefaultRtspTcpPort         = "8554"
+	RtspAuthenticationServer        = "RtspAuthenticationServer"
+	DefaultRtspAuthenticationServer = "localhost:8000"
```
ajcasagrande, line 24: this doesn't look go formatted

internal/driver/driver.go
```diff
@@ -48,20 +48,22 @@ const (
 )
 
 const (
+	// Key used in the secret store for RTSP credentials.
```
ajcasagrande, line 51: start comment with the name, like so
```suggestion
	// rtspAuthKey defines the secretName used for storing RTSP credentials in the secret store.
```

internal/driver/driver.go
```diff
@@ -155,47 +164,45 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
-		d.lc.Warnf("could not unmarshal read body: %s\n", err)
-		w.WriteHeader(http.StatusUnauthorized)
+		d.lc.Errorf("could not unmarshal read body: %s", err)
+		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
 
+	if rtspAuthRequest.User == "" || rtspAuthRequest.Password == "" {
+		d.lc.Debug("username or password is empty") // this can happen during normal operation
+		w.WriteHeader(http.StatusBadRequest)
```
ajcasagrande, line 191: No, this needs to return 401 Unauthorized as per the mediamtx documentation.

internal/driver/driver.go
```diff
@@ -693,11 +693,17 @@ func (d *Driver) newDevice(name string, protocols map[string]models.ProtocolProp
...
+		Scheme: RtspUriScheme,
+		Host:   fmt.Sprintf("%s:%s", d.rtspHostName, d.rtspTcpPort),
+	}
+	rtspUriNoCredentials.Path = path.Join(Stream, name)
 	return &Device{
 		name:                        name,
 		path:                        fdPath,
 		serialNumber:                sn,
 		rtspUri:                     rtspUri.String(),
+		rtspUriNoCredentials:        rtspUriNoCredentials.String(),
```
ajcasagrande, line 706: I do not think you need to store an additional variable here. the regular rtspUri should be unauthenticated, and only the authenticated one is created on the fly and not stored on the device struct.

For example, create the authenticated one separatly, and use it to init the trans: `err = trans.Initialize(fdPath, rtspAuthenticatedUri.String())`

ajcasagrande (changes requested):
Missing all documentation.

ajcasagrande:
This PR needs to be based on `main` branch and not `levski`.

internal/driver/driver.go
```diff
@@ -147,9 +160,52 @@ func (d *Driver) Initialize(lc logger.LoggingClient, asyncCh chan<- *sdkModels.A
...
+	}
+
+	if rtspAuthRequest.User == "" || rtspAuthRequest.Password == "" {
+		d.lc.Debug("username or password is empty") // this can happen during normal operation
+		w.WriteHeader(http.StatusBadRequest)
+	}
+	credential, edgexErr := d.tryGetCredentials(rtspAuthKey)
+	if edgexErr != nil {
+		d.lc.Warnf("Failed to retrieve credentials for rtsp authentication from the secret store. Have you stored credentials yet for secretName %s?", rtspAuthKey)
+		w.WriteHeader(http.StatusUnauthorized)
```
ajcasagrande, line 196: I think we should write in the response body that there is a problem with the configuration. this could help user debug it if they forget to set it up.

internal/driver/credentials.go
ajcasagrande: still needs to be addressed. delete this block of code, import `"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/secret"`, and then use `secret.UsernameKey` below.

internal/driver/driver.go
```diff
@@ -562,6 +619,13 @@ func (d *Driver) newDevice(name string, protocols map[string]models.ProtocolProp
 		Scheme: RtspUriScheme,
 		Host:   fmt.Sprintf("%s:%s", d.rtspHostName, d.rtspTcpPort),
 	}
+	credential, edgexErr := d.tryGetCredentials(rtspAuthKey)
+	if edgexErr != nil {
+		d.lc.Warnf("Failed to get credentials for rtsp authentication from secretName %s", rtspAuthKey)
```
ajcasagrande, line 624: I would add in the log that authentication with the rtsp server will be disabled.

PR #240, opened 2023-05-08
feat: support for rtsp server authentication

brianointel:

## PR Checklist

- [ ] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [ ] I am not introducing a new dependency (add notes below if you are)
  <link to docs PR>


internal/driver/driver.go
```diff
@@ -121,12 +133,51 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
...
+	}
+
+	if rtspAuthRequest.User == "" || rtspAuthRequest.Password == "" {
+		d.lc.Debug("username or password is empty") // this can happen during normal operation
+		w.WriteHeader(http.StatusBadRequest)
+	}
+	credential, edgexErr := d.tryGetCredentials(rtspAuthKey)
+	if edgexErr != nil {
+		d.lc.Warnf("Failed to retrieve credentials for rtsp authentication from the secret store. Have you stored credentials yet for secretName %s?", rtspAuthKey)
+		w.WriteHeader(http.StatusUnauthorized)
```
ajcasagrande: I think we should write in the response body that there is a problem with the configuration. this could help user debug it if they forget to set it up.

internal/driver/driver.go
```diff
@@ -459,6 +511,13 @@ func (d *Driver) newDevice(name string, protocols map[string]models.ProtocolProp
 		Scheme: RtspUriScheme,
 		Host:   fmt.Sprintf("%s:%s", d.rtspHostName, d.rtspTcpPort),
 	}
+	credential, edgexErr := d.tryGetCredentials(rtspAuthKey)
+	if edgexErr != nil {
+		d.lc.Warnf("Failed to get credentials for rtsp authentication from secretName %s", rtspAuthKey)
```
ajcasagrande: I would switch to using the wording from the other log:

```suggestion
		d.lc.Warnf("Failed to retrieve credentials for rtsp authentication from the secret store. Have you stored credentials yet for secretName %s?", rtspAuthKey)
```

internal/driver/types.go
```diff
@@ -0,0 +1,18 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+//
+// Copyright (C) 2023 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package driver
+
+type RTSPAuthRequest struct {
+	Ip       string `json:"ip"`
```
ajcasagrande: should be `IP`

internal/driver/types.go
```diff
@@ -0,0 +1,18 @@
...
+	User     string `json:"user"`
+	Password string `json:"password"`
+	Path     string `json:"path"`
+	Protocol string `json:"protocol"`
+	Id       string `json:"id"`
```
ajcasagrande: Should be `ID`

internal/driver/driver.go
```diff
@@ -526,11 +585,17 @@ func (d *Driver) newDevice(name string, protocols map[string]models.ProtocolProp
 			fmt.Sprintf("wrong device serial number, expected %s=%s, actual %s=%s", SerialNumber, psn, SerialNumber, sn), nil)
 	}
 
+	rtspUriNoCredentials := &url.URL{
+		Scheme: RtspUriScheme,
+		Host:   fmt.Sprintf("%s:%s", d.rtspHostName, d.rtspTcpPort),
+	}
+	rtspUriNoCredentials.Path = path.Join(Stream, name)
```
ajcasagrande: I do not think you need to store an additional variable here. the regular rtspUri should be unauthenticated, and only the authenticated one is created on the fly and not stored on the device struct.

For example, create the authenticated one separatly, and use it to init the trans: `err = trans.Initialize(fdPath, rtspAuthenticatedUri.String())`

internal/driver/driver.go
```diff
@@ -121,12 +133,51 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
...
+	err = json.Unmarshal(body, &rtspAuthRequest)
+	if err != nil {
+		d.lc.Errorf("could not unmarshal read body: %s", err)
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	if rtspAuthRequest.User == "" || rtspAuthRequest.Password == "" {
+		d.lc.Debug("username or password is empty") // this can happen during normal operation
+		w.WriteHeader(http.StatusBadRequest)
```
ajcasagrande: No, this needs to return 401 Unauthorized as per the mediamtx documentation.

internal/driver/constants.go
```diff
@@ -20,6 +20,8 @@ const (
 	DefaultRtspServerHostName = "localhost"
 	RtspTcpPort               = "RtspTcpPort"
 	DefaultRtspTcpPort        = "8554"
+	RtspAuthenticationServer        = "RtspAuthenticationServer"
+	DefaultRtspAuthenticationServer = "localhost:8000"
```
ajcasagrande, line 24: run go fmt

go.mod
```diff
@@ -55,6 +55,7 @@ require (
 	github.com/nats-io/nats.go v1.25.0 // indirect
 	github.com/nats-io/nkeys v0.4.4 // indirect
 	github.com/nats-io/nuid v1.0.1 // indirect
+	github.com/pkg/errors v0.9.1 // indirect
```
ajcasagrande: hmm, what is bringing in this? Run a `go mod tidy` and make sure it still belongs.

cmd/res/configuration.yaml
```diff
@@ -34,3 +34,4 @@ Device:
 Driver:
   RtspServerHostName: "localhost"
   RtspTcpPort: "8554"
+  RtspAuthenticationServer = "localhost:8000"
```
ajcasagrande:
```suggestion
  RtspAuthenticationServer: "localhost:8000"
```

Dockerfile
```diff
@@ -59,6 +59,7 @@ COPY --from=rtsp /rtsp-simple-server /
 RUN sed -i 's/rtmpDisable: no/rtmpDisable: yes/g' rtsp-simple-server.yml
 RUN sed -i 's/hlsDisable: no/hlsDisable: yes/g' rtsp-simple-server.yml
 RUN sed -i 's/protocols: \[udp, multicast, tcp\]/protocols: \[tcp\]/g' rtsp-simple-server.yml
+RUN sed -i 's,externalAuthenticationURL:,externalAuthenticationURL: http://localhost:8000/rtspauth,g' rtsp-simple-server.yml
```
ajcasagrande, line 62: We need to check and see if this can be overriden via env variables or not.

internal/driver/driver.go
```diff
@@ -34,6 +34,10 @@ import (
 var driver *Driver
 var once sync.Once
 
+const (
+	// Key used in the secret store for RTSP credentials.
```
```suggestion
	// rtspAuthKey defines the secretName used for storing RTSP credentials in the secret store.
```

internal/driver/credentials.go
```diff
@@ -0,0 +1,39 @@
...
+// Credentials encapsulates Username and Password pair.
+type Credentials struct {
+	Username string
+	Password string
+}
+
+const (
+	UsernameKey = "username"
+	PasswordKey = "password"
+)
```
ajcasagrande: still needs to be addressed. delete this block of code, import `"github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/secret"`, and then use `secret.UsernameKey` below.

internal/driver/driver.go
```diff
@@ -121,12 +133,51 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
...
 	go d.RefreshExistingDevicePaths()
-
+	go d.StartRtspCredentialServer()
 	return nil
 }
+func (d *Driver) StartRtspCredentialServer() {
+	d.lc.Infof("starting rtsp server")
+	rtspAuthServer := http.NewServeMux()
+	rtspAuthServer.HandleFunc("/rtspauth", d.RTSPCredentialsHandler)
+	http.ListenAndServe(d.rtspAuthenticationServer, rtspAuthServer)
```
ajcasagrande: I think we probably need to add code to Stop the server on `Stop()`

internal/driver/driver.go
```diff
@@ -121,12 +133,51 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
 
 	// Make sure the paths of existing devices are up-to-date.
 	go d.RefreshExistingDevicePaths()
-
+	go d.StartRtspCredentialServer()
 	return nil
 }
+func (d *Driver) StartRtspCredentialServer() {
```
```suggestion
func (d *Driver) StartRTSPCredentialServer() {
```

ajcasagrande (changes requested):
Need to add documentation on what the feature is, how the feature works, and how to set the credentials in the secret store.

ajcasagrande (changes requested):
Need to add documentation on what the feature is, how the feature works, and how to set the credentials in the secret store.

ajcasagrande:
@brianointel Please fill out the PR Checklist

go.mod
brianointel: it's used in credentials.go

internal/driver/driver.go
brianointel: should i do something like this https://github.com/rudderlabs/graceful-shutdown-examples/blob/main/httpserver/main.go

ajcasagrande: @brianointel A little similar, but I think it can be done cleaner by adding a waitgroup and a pointer to the http server in the driver struct. In initialize(), before calling `go StartRTSP...` function, do rtspAuthWaitGroup.Add(1), and as the first line of this StartRTSP... func, do `defer rtspAuthWaitGroup.Done()`. Then in the driver.Stop() function, shutdown the server, and then call `rtspAuthWaitGroup.Wait()`.

go.mod
ajcasagrande: @brianointel No, that uses `github.com/edgexfoundry/go-mod-core-contracts/v2/errors`, and if you check that package, it is not listed as a dependency either. https://github.com/edgexfoundry/go-mod-core-contracts/blob/main/go.mod.

go.mod
```diff
@@ -4,21 +4,31 @@ go 1.20
...
 require (
 	github.com/Microsoft/go-winio v0.6.0 // indirect
 	github.com/OneOfOne/xxhash v1.2.8 // indirect
 	github.com/armon/go-metrics v0.3.10 // indirect
 	github.com/cenkalti/backoff v2.2.1+incompatible // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/eclipse/paho.mqtt.golang v1.4.2 // indirect
-	github.com/edgexfoundry/go-mod-bootstrap/v3 v3.0.0-dev.80 // indirect
+	github.com/edgexfoundry/go-mod-bootstrap v0.0.72
+	github.com/edgexfoundry/go-mod-bootstrap/v3 v3.0.0-dev.81 // indirect
```
ajcasagrande: there is something off about this go.mod file. why is it using both v0 and v3 go-mod-bootstrap? Also, there should only be 2 sections and not 3.

lenny-goodell: > there is something off about this go.mod file. why is it using both v0 and v3 go-mod-bootstrap? Also, there should only be 2 sections and not 3.

Wow, that is interesting. @brianointel you need to clear the v0.0.72 from your module cache and find the import that is referencing `github.com/edgexfoundry/go-mod-bootstrap` and fix it to be `github.com/edgexfoundry/go-mod-bootstrap/v3`

internal/driver/credentials.go
```diff
@@ -0,0 +1,35 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+//
+// Copyright (C) 2023 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package driver
+
+import (
+	"github.com/edgexfoundry/go-mod-bootstrap/bootstrap/secret"
```
ajcasagrande: @brianointel this is where the bad import is coming from. needs to be updated to v3

ajcasagrande (dismissed):
LGTM

ajcasagrande (approved):

Issue #23, opened 2022-05-19
StartStreaming errors not passed to the caller


farshidtz:
# 🐞 Bug Report

### Affected Services [**REQUIRED**]

 The issue is located in: 

### Is this a regression?

 No

### Description and Minimal Reproduction [**REQUIRED**]

Make sure that the rtsp is NOT running.

Start streaming:
```
$ curl -X PUT -d '{                                                                              
    "StartStreaming": {
      "InputImageSize": "640x480",
      "OutputVideoQuality": "5"
    }
}' http://localhost:59882/api/v2/device/name/Logitech_StreamCam-4B17BE45/StartStreaming
{"apiVersion":"v2","statusCode":200}
```

This returns 200 OK as if the requested operation was successful. If this HTTP request is just meant to accept the the operation and run it asynchornously, the response code should be [202 Accepted](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202)

Logs:
```
level=INFO ts=2022-05-19T11:48:31.864475627Z app=device-usb-camera source=driver.go:536 msg="start video streaming for device Logitech_StreamCam-4B17BE45"
level=ERROR ts=2022-05-19T11:48:32.774545384Z app=device-usb-camera source=driver.go:524 msg="the video streaming process for device Failed Finish FFMPEG ([-nostats -loglevel 0 -y -s 640x480 -i /dev/video0 -qscale 5 -f rtsp -hls_list_size 0 rtsp://localhost:8554/stream/Logitech_StreamCam-4B17BE45]) with exit status 1 message   has stopped, error: %!s(MISSING)"
```

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍  Your Environment
**Deployment Environment:**

**EdgeX Version [**REQUIRED**]:**

**Anything else relevant?**

lenny-goodell:
@farshidtz , did  #24 address this issue?
@cloudxxx8 , if issue still needs addressing, can your team take it?

farshidtz:
> @farshidtz , did  #24 address this issue?

No, that is a fix for another issue.

cloudxxx8:
@farshidtz this is the normal command API, and it is a **Set Command** provided by this Device Service
Set Command only supports 200 response for now.
This should be applied to the new device function in the future
https://github.com/edgexfoundry/edgex-docs/pull/644

lenny-goodell:
[...]

PR #254, opened 2023-05-19
fix: return ffmpeg error logs to caller, and fix StreamingStatus

### LINKED ISSUE #253: StartStreaming fails with unknown error when rtsp auth credentials are not set
# 🐞 Bug Report

When the rtspauth credentials have not been set, calling StartStreaming will fail due to credential issues (as intended), however the caller of the StartStreaming command does not know why this happened.

## 🌍  Your Environment
**Deployment Environment:**
hybrid or full docker-compose

**EdgeX Version [**REQUIRED**]:**
latest main

**Anything else relevant?**

We should return to the caller something that conveys there is an issue with the credentials.

ajcasagrande:
FFmpeg transcoder logic has been implemented within the driver code now, so that way we have more control over it. This allows us to keep track of the process output and log it to the console as well as return it to the caller of StartStreaming. Because we have more precise control over the process, we can better sync the StreamingStatus to whether or not the ffmpeg process is running.

The handling of publishing StreamingStatus to the message bus has been optimized to send better based on when it actually changes, and not send when the user calls StopStreaming when already stopped, or StartStreaming when already started.

Logic has also been added to redact the rtsp credentials from the log file and API responses to avoid exposing secrets to the caller.

## PR Checklist

- [X] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [X] I am not introducing a new dependency (add notes below if you are)
- [X] I have fully tested (add details below) this the new feature or bug fix (if not, why?)
- [ ] I have opened a PR for the related docs change (if not, why?) **N/A**
  <link to docs PR>

## Testing Instructions

- checkout and build this PR
- run edgex in secure or non secure mode using the image built from this pr
  - update `add-device-usb-camera.yaml`
  - run `make run ds-usb-camera`
- wait for your camera to be auto-detected, and get the device name
- execute StartStreaming, and expect a 500 error with a log message that contains messages related to 401 authentication error
- execute StreamingStatus and verify isstreaming=false and error is the same (or a subset) of the api call result
- post your rtspauth credentials to `/api/v3/secrets`  using the guide from edgex-docs (see **step 4** [here](https://docs.edgexfoundry.org/3.0/microservices/device/supported/device-usb-camera/walkthrough/deployment/#verify-service-device-profiles-and-device))
- execute StartStreaming with **valid** inputs and expect it to return a 200 OK, taking about 1 second as before
- view the stream using mplayer or ffplay (see [here](https://docs.edgexfoundry.org/3.0/microservices/device/supported/device-usb-camera/walkthrough/general-usage/))
- execute StreamingStatus and see that there is no error, and isstreaming=true
- execute StartStreaming and expect an error already running
- execute StreamingStatus and see that there is no error, and isstreaming=true
- execute StopStreaming and expect 200 OK
- execute StreamingStatus and see that there is no error, and isstreaming=false
- execute StopStreaming and expect 200 OK
- execute StreamingStatus and see that there is no error, and isstreaming=false
- execute StartStreaming with **invalid** inputs and expect it to return a 500 error, with details about the error such as `mpeg456` not being valid
- execute StreamingStatus and verify isstreaming=false and error is the same (or a subset) of the api call result
- execute StopStreaming and expect 200 OK
- execute StreamingStatus and nothing has changed

additional things to validate:
- start streaming and then stop the device service while still streaming. make sure the camera is no longer active. start the service back up and ensure it starts fine and u can start streaming the same camera again no problems.
- try and start streaming invalid input, and then make sure stream with valid afterwards still works
- try and start streaming with invalid input and then stop the device service. start the service back up and ensure it starts fine and u can start streaming the same camera again no problems.

## New Dependency Instructions (If applicable)

Fixes #253 
Implements a proper fix for #23

### DIFF
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -676,32 +676,52 @@ func (d *Driver) getDevice(name string) (*Device, errors.EdgeX) {
 }
 
 func (d *Driver) startStreaming(device *Device) errors.EdgeX {
-	errChan, err := device.StartStreaming()
+	progressChan, errChan, err := device.StartStreaming()
 	if err != nil {
 		return errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf(
 			"failed to start video streaming for device %s", device.name), err)
 	}
 
 	defer func() {
+		// before we return, publish the current streaming status right away.
+		// do this even on error, as this will contain the error message as well.
 		go d.publishStreamingStatus(device)
 	}()
 
+	waitForFinishAndPublish := func() {
+		d.lc.Debugf("Waiting for ffmpeg errChan to be done")
+		<-errChan
+		d.lc.Debugf("Done waiting for ffmpeg errChan to be done")
+		d.publishStreamingStatus(device)
+	}
+
 	// wait a little bit before returning to see if there are any errors on startup
-	select {
-	case <-time.After(time.Second):
-		d.lc.Infof("Video streaming for device %s has started without error", device.name)
-		go func() {
-			// wait for the process to complete in the background and then publish the streaming status
-			<-errChan
-			d.publishStreamingStatus(device)
-		}()
-		return nil
-	case startErr := <-errChan:
-		if startErr == nil {
+	ticker := time.NewTicker(4 * time.Second)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+			// this should rarely happen, as ffmpeg should print progress on the first frame. If it does happen,
+			// then either progress has been disabled or the process could be having issues.
+			d.lc.Warnf("Video streaming for device %s has started but has not sent progress messages yet.", device.name)
+			go waitForFinishAndPublish() // track process in the background
+			return nil
+		case startErr, ok := <-errChan:
+			if startErr == nil || !ok {
+				d.lc.Warnf("Video streaming for device %s seems to have stopped already.", device.name)
+				return nil
+			}
+			return errors.NewCommonEdgeX(errors.KindServerError,
+				fmt.Sprintf("the video streaming process for device %s has stopped", device.name), startErr)
+		case _, ok := <-progressChan:
+			if !ok {
+				continue // channel was closed, so something else must be the problem
+			}
+			// if we got a progress message, that means that the transcoding is successful
+			d.lc.Infof("Video streaming for device %s has started without error", device.name)
+			go waitForFinishAndPublish() // track process in the background
 			return nil
 		}
-		return errors.NewCommonEdgeX(errors.KindServerError,
-			fmt.Sprintf("the video streaming process for device %s has stopped", device.name), startErr)
 	}
 }
 
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -153,7 +153,7 @@ func (d *Driver) Start() error {
 }
 
 func (d *Driver) StartRTSPCredentialServer() {
-	d.lc.Infof("Starting rtsp server")
+	d.lc.Infof("Starting rtsp authentication server on %s", d.rtspAuthenticationServerUri)
 	defer d.wg.Done()
 
 	router := mux.NewRouter()
@@ -347,11 +347,10 @@ func (d *Driver) HandleWriteCommands(deviceName string, protocols map[string]mod
 				return errors.NewCommonEdgeXWrapper(edgexErr)
 			}
 		case VideoStopStreaming:
-			device.StopStreaming(nil)
+			device.StopStreaming()
 		default:
 			return errors.NewCommonEdgeX(errors.KindContractInvalid, fmt.Sprintf("unsupported command %s", command), nil)
 		}
-		go d.publishStreamingStatus(device)
 	}
 
 	return nil
@@ -376,7 +375,7 @@ func (d *Driver) Stop(force bool) error {
 
 	for _, device := range d.activeDevices {
 		go func(device *Device) {
-			device.StopStreaming(nil)
+			device.StopStreaming()
 			d.wg.Done()
 		}(device)
 	}
@@ -452,7 +451,7 @@ func (d *Driver) RemoveDevice(deviceName string, protocols map[string]models.Pro
 	d.mutex.Lock()
 	defer d.mutex.Unlock()
 	if device, ok := d.activeDevices[deviceName]; ok {
-		device.StopStreaming(nil)
+		device.StopStreaming()
 		delete(d.activeDevices, deviceName)
 		d.lc.Debugf("Device %s is removed", deviceName)
 	}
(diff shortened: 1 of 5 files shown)


ChristianDarr-personal:
To start and view the stream, see [here](https://docs.edgexfoundry.org/3.0/microservices/device/supported/device-usb-camera/walkthrough/general-usage/)

For how to set credentials - see step 4 [here](https://docs.edgexfoundry.org/3.0/microservices/device/supported/device-usb-camera/walkthrough/deployment/#verify-service-device-profiles-and-device)

ajcasagrande:
It looks like `gosec` is flagging a `G204: Subprocess launched with a potential tainted input or cmd arguments (gosec)`.

This should be evaluated, however do understand that the `G204` already exists in the code base and is just not being detected due to the fact it resides in a 3rd-party library.

**Update:**
After researching, the way it is utilized is safe. The reason G204 was being flagged was because `exec.Command` was being passed a function, and not a variable. Since the function call is just a getter for a variable, I moved it out into a temp var and that removed the warning. 

More info: 

https://github.com/securego/gosec/blob/87cc45e1cd903e2038e868eaf026cbc5d1dd1a26/rules/subproc.go#L34-L42

https://github.com/securego/gosec/blob/87cc45e1cd903e2038e868eaf026cbc5d1dd1a26/rules/subproc.go#L91C89-L92

internal/driver/transcoder.go
```diff
@@ -0,0 +1,179 @@
...
+
+	// in the background we will wait for the process to complete and return any errors over the done channel
+	done := make(chan error)
+	go func() {
+		defer close(done)
+
+		// wait until the process has exited
+		err = proc.Wait()
+		dev.lc.Debugf("FFmpeg process with pid %d for device %s exited with code %d. User time: %v, System time: %v",
+			proc.Process.Pid, dev.name, proc.ProcessState.ExitCode(), proc.ProcessState.UserTime(), proc.ProcessState.UserTime())
```
FelixTing:
```suggestion
		dev.lc.Debugf("FFmpeg process with pid %d for device %s exited with code %d. User time: %v, System time: %v",
			proc.Process.Pid, dev.name, proc.ProcessState.ExitCode(), proc.ProcessState.UserTime(), proc.ProcessState.SystemTime())
```

FelixTing (changes requested):

vyshali-chitikeshi:
I validated this PR  thoroughly both in secure and non-secure modes.
Tested what was recommend in test instructions of this PR and I added some of my own tests.
Functionality wise it works correctly, no issues seen. Tried several time start steaming with valid and invalid inputs, stop streaming, restarting/stopping/starting usb service while streaming is going on, etc checked streaming status for every test to ensure it reflects correct streaming status.
But it gives status code 200 in below scenarios:
Start streaming with valid inputs without rtsp creds uploaded
Start streaming with invalid inputs
Based on PR testing instructions, its supposed to give status code 500 for above 2 cases.
However though it gives status code 200 in above 2 cases, streaming will not start and gives appropriate error in streaming status api error section. only problem is api output status code. Find attached logs and screenshots.

vyshali-chitikeshi:
streaming_status_with_invalid_inputs
start_streaming_with_invalid_inputs
streaming_status_without_rtsp_creds_with_valid_inputs
start_streaming_without_rtsp_credswith_valid_inputs
[logs_without_uploading_rtsp_creds.txt](https://github.com/edgexfoundry/device-usb-camera/files/11536609/logs_without_uploading_rtsp_creds.txt)

internal/driver/driver.go
```diff
@@ -676,41 +676,32 @@ func (d *Driver) getDevice(name string) (*Device, errors.EdgeX) {
...
+		go d.publishStreamingStatus(device)
 	}()
-	for {
-		select {
-		case <-time.After(time.Second):
-			d.lc.Infof("start video streaming for device %s", device.name)
+
+	// wait a little bit before returning to see if there are any errors on startup
+	select {
+	case <-time.After(time.Second):
```
FelixTing: Sometimes, one second is not enough for FFmpeg to complete the execution of the start streaming command. This could be the root cause of the issue @vyshali-chitikeshi encountered.
I tried extending it to 3 seconds, and then I was able to receive the expected 500 status code when the streaming failed to start.

However, defining how long to wait seems challenging as it may depend on machine performance. We also need to consider the HTTP request timeout, which is 5 seconds by default. https://github.com/edgexfoundry/edgex-go/blob/3d7959681c73ad1dd76784b2336248bd6b6384c3/cmd/core-common-config-bootstrapper/res/configuration.yaml#L32

ChristianDarr-personal (commented):
LGTM

internal/driver/driver.go
ajcasagrande: Yeah, originally I had made this to be 2 seconds (instead of the current 1 second) when I started this work, but I dropped it back down to 1 as I noticed my typical error response was like 400-600ms. However, I think that may be based on what the actual error was (eg. bad argument, rtsp server offline, etc), and like you said, system performance. In theory we could make this configurable as well.

I had an idea that we could use something in the output stream to determine if the streaming has started ok, but was not sure what we could parse as a potential indicator of that. I know enabling progress logs would allow that, but doesn't seem worth it for just grabbing beginning of stream confirmation.

The last idea is that we return right away, and expect the user to monitor StreamingStatus endpoint for any errors, since as @vyshali-chitikeshi has validated, that always contains the error when the api returned success too early.

ajcasagrande:
@vyshali-chitikeshi @FelixTing I have just pushed updated code to handle this situation. It now waits up to 4 seconds before returning in case the error takes longer on certain machines. I have also enabled ffmpeg progress reporting of 60 second intervals. This way it is not super chatty, but still allows us to have something to parse when the stream has successfully started.

In my experimentation, I have noticed ffmpeg will always print progress for the first frame, so no matter if the reporting interval is 1 minute, we still get notified right away. This seems to be the best of both worlds, as it allows more time for slower machines/commands, but doesn't drastically delay the response of successful streams.

_Successful response in 669ms_
Screenshot from 2023-05-24 16-32-02

_Error response in 1097ms_
Screenshot from 2023-05-24 16-32-39

vyshali-chitikeshi:
With the updated code, it works correctly. Its gives status code 500 with appropriate error message n below 2 scenarios as expected
Start streaming with valid inputs without uploading rtsp creds
Start streaming with invalid inputs
attached screenshots.
Screenshot from 2023-05-24 23-51-02
Screenshot from 2023-05-24 23-50-14


Also tested other api's, and some negative scenarios like starting streaming while streaming in-progress and restarting usb service while streaming is going on etc Everything looks good.

vyshali-chitikeshi:
Also from performance perspective, I don't see noticeable delay to get rest-api outputs.

FelixTing (approved):
LGTM

FelixTing (approved):
LGTM

PR #261, opened 2023-06-16
fix: panic when device is missing CardName or SerialNumber

### LINKED ISSUE #263: Panic when device is missing CardName or SerialNumber (Minnesota)
# 🐞 Bug Report

device-usb-camera

### Is this a regression?
No

- Run 3.0 version of Edgex core services along with device-usb-camera service (with discovery disabled)
- Add usb camera manually using REST endpoint and omit CardName and/or SerialNumber info while adding. E.g.
      
```
curl -X POST -H 'Content-Type: application/json'  http://localhost:59881/api/v2/device -d '[
    {
    "apiVersion" : "v2",
    "device": {
        "name": "usb-camera",
        "serviceName": "device-usb-camera",
        "profileName": "USB-Camera-General",
        "description": "My test camera",
        "adminState": "UNLOCKED",
        "operatingState": "UP",
        "protocols": {
            "USB": {
                "CardName": "Webcam C270 HD",
                "Path": "/dev/video0",
                "AutoStreaming": "false"
            }
        }
    }
    }
]'
```

- The device is successfully added into Edgex but when we execute some command on the device such as `StartStreaming` the device-usb-camera service panics with the error `panic: interface conversion: interface {} is nil, not string` and eventually exits.

## 🔥 Exception or Error
`panic: interface conversion: interface {} is nil, not string
`
## 🌍  Your Environment
**Deployment Environment:** Docker, Docker + Dev Hybrid

**EdgeX Version [**REQUIRED**]:** 3.0

**Anything else relevant?**
This issue will be fixed in `main`. `CardName` and `SerialNumber` are optional parameters but the service does not have proper checks and when the device is added manually without one or both of these parameters the issue is seen. With the fixed service, based on their respective needs users can manually add a camera with or without these parameters.

ajcasagrande:
See: https://github.com/orgs/edgexfoundry/discussions/129

## PR Checklist

- [X] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [X] I am not introducing a new dependency (add notes below if you are)
- [N/A] I have added unit tests for the new feature or bug fix (if not, why?)
  <link to docs PR>

### DIFF
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -11,6 +11,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/edgexfoundry/go-mod-core-contracts/v3/dtos"
 	"github.com/gorilla/mux"
 	"io"
 	"net/http"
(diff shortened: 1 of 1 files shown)


lenny-goodell:
@ajcasagrande , please fix on main (not minnesota since there is a work around) and create a BUG issue for it and list it here: https://wiki.edgexfoundry.org/display/FA/Minnesota#Minnesota-KnownBugs

presatish (commented):
It would be helpful if examples under https://github.com/EdgeX-Camera-Management/device-usb-camera/blob/fix-discover-panic/cmd/res/devices are updated too. There is no `CardName` here.

internal/driver/driver.go
```diff
@@ -747,9 +747,9 @@ func (d *Driver) cachedDeviceMap() map[string]models.Device {
 	cds := d.ds.Devices()
 	cdm := make(map[string]models.Device, len(cds))
 	for _, cd := range cds {
-		cn := cd.Protocols[UsbProtocol][CardName].(string)
-		sn := cd.Protocols[UsbProtocol][SerialNumber].(string)
-		if len(cn) > 0 && len(sn) > 0 {
+		cn, cnOK := cd.Protocols[UsbProtocol][CardName].(string)
```
lenny-goodell, line 766: Isn't it the .(string) that is causing the panic if the item is not in the map?

lenny-goodell (changes requested):

internal/driver/driver.go
ajcasagrande: Yes, and the `.(string)` should also be the one returning the cnOK

lenny-goodell (approved):
LGTM

lenny-goodell (changes requested):
> It would be helpful if examples under https://github.com/EdgeX-Camera-Management/device-usb-camera/blob/fix-discover-panic/cmd/res/devices are updated too. There is no `CardName` here.

Yes, and not just in the Example App, also here in this repo.
https://github.com/edgexfoundry/device-usb-camera/blob/main/cmd/res/devices/general.usb.camera.yaml.example
https://github.com/edgexfoundry/device-usb-camera/blob/main/cmd/res/devices/hp.w200.yaml.example

@ajcasagrande , please fix these as part of this PR. THX!

internal/driver/driver.go
```diff
@@ -606,23 +607,37 @@ func (d *Driver) newDevice(name string, protocols map[string]models.ProtocolProp
...
 			fmt.Sprintf("could not find the serial number of the device on the specified path: %s", fdPath), err)
 	}
-	psn := protocols[UsbProtocol][SerialNumber].(string)
-	pcn := protocols[UsbProtocol][CardName].(string)
-	// pre-defined devices may not include serial number information
-	if len(psn) == 0 {
+
+	// if the user provided a serial number or card name, but it does not match the device's serial number or card name,
+	// then return an error, as this may not be the correct device
+	psn, psnOK := protocols[UsbProtocol][SerialNumber].(string)
```
lenny-goodell, line 613: Should also validate `protocols[UsbProtocol]` exists before index on it.

lenny-goodell (changes requested):

internal/driver/driver.go
ajcasagrande: That should already be handled by the validateDevice call I believe. So I wouldn't expect it to be able to reach here unless it had that.

lenny-goodell: Shouldn't validateDevice also ensure SerialNumber and CardName are present?
Or are they somehow optional?

presatish:
> > It would be helpful if examples under https://github.com/EdgeX-Camera-Management/device-usb-camera/blob/fix-discover-panic/cmd/res/devices are updated too. There is no `CardName` here.
> 
> Yes, and not just in the Example App, also here in this repo. https://github.com/edgexfoundry/device-usb-camera/blob/main/cmd/res/devices/general.usb.camera.yaml.example https://github.com/edgexfoundry/device-usb-camera/blob/main/cmd/res/devices/hp.w200.yaml.example
> 
> @ajcasagrande , please fix these as part of this PR. THX!

@lenny-intel After looking into the issue more I don't think `cardName` and `serialNumber` need to be included in the examples. They are not the required parameters, they can be used if additional checking is required. Initially when I posted my comment I thought they are the required parameters.

internal/driver/driver.go
presatish: `SerialNumber` and `CardName` are optional parameters.

lenny-goodell (approved):
LGTM

lenny-goodell (changes requested):
Still need to update the example device files to have the work around

lenny-goodell (approved):
LGTM, ok with not updating the examples since the extras are optional.

PR #265, opened 2023-07-13
feat: SetFps v4l2 + ffmpeg command

ChristianDarr-personal:

## PR Checklist

  <link to docs PR>

## Testing Instructions

Note, any commands executed need to include the query parameter `PathIndex` with the value set to the number representing the video path stream intended. For example: `http://localhost:59982/api/v3/url?PathIndex=0`. Also it is recommended to use the updated postman collection.
1. Execute the `FrameRateFormat` api to see supported fps values. This shows all fps values for each resolution for each type of encoding. This can be useful for determining the values to set the camera fps.
2. Execute the `DataFormat` api to see the supported fps values for the current video format and encoding.
3. Execute the `SetFrameRate` api and see no errors.The fps is represented in a time per frame format, meaning the denominator divided by the numerator yields the fps value. I did this to maintain consistency with the internal data structures. You can leave the numerator out completely, and the denominator will represent the fps value.
4. Execute the `GetFrameRate` api and see that the fps match.
5. Test the video stream and/or raw device stream to see if the changes apply. Certain video players seem to reset the fps value themselves, so it may not be viable to see the fps value of the /dev/video stream reflected in them.

### DIFF
--- a/cmd/res/profiles/general.usb.camera.yaml
+++ b/cmd/res/profiles/general.usb.camera.yaml
@@ -90,7 +90,7 @@ deviceResources:
   - name: "SetFramerate"
     description: "Set the stream Framerate rate"
     attributes: 
-      { command: "VIDEO_SET_FPS"}
+      { command: "VIDEO_SET_FRAMERATE"}
     properties:
       valueType: "Object"
       readWrite: "W"
--- a/cmd/res/profiles/general.usb.camera.yaml
+++ b/cmd/res/profiles/general.usb.camera.yaml
@@ -44,6 +44,13 @@ deviceResources:
     properties:
       valueType: "Object"
       readWrite: "R"
+  - name: "FpsFormats"
+    description: "Enumerate supported fps settings, see https://linuxtv.org/downloads/v4l-dvb-apis/userspace-api/v4l/vidioc-enum-frameintervals.html"
+    attributes:
+      { command: "METADATA_FPS_FORMATS" }
+    properties:
+      valueType: "Object"
+      readWrite: "R"
   - name: "DataFormat"
     description: "Get data format, see https://linuxtv.org/downloads/v4l-dvb-apis/userspace-api/v4l/vidioc-enuminput.html."
     attributes:
@@ -80,7 +87,7 @@ deviceResources:
       valueType: "Bool"
       readWrite: "W"
       defaultValue: "false"
-  - name: "SetFPS"
+  - name: "SetFps"
     description: "Set the stream FPS rate"
     attributes: 
       { command: "VIDEO_SET_FPS"}
--- a/cmd/res/profiles/general.usb.camera.yaml
+++ b/cmd/res/profiles/general.usb.camera.yaml
@@ -80,6 +80,13 @@ deviceResources:
       valueType: "Bool"
       readWrite: "W"
       defaultValue: "false"
+  - name: "SetFPS"
+    description: "Set the stream FPS rate"
+    attributes: 
+      { command: "VIDEO_SET_FPS"}
+    properties:
+      valueType: "Object"
+      readWrite: "W"
   - name: "StreamURI"
     description: "Get video-streaming URI."
     attributes:
--- a/cmd/res/profiles/general.usb.camera.yaml
+++ b/cmd/res/profiles/general.usb.camera.yaml
@@ -94,6 +94,13 @@ deviceResources:
     properties:
       valueType: "Object"
       readWrite: "W"
+  - name: "GetFramerate"
+    description: "Get the stream Framerate rate"
+    attributes: 
+      { command: "VIDEO_GET_FRAMERATE"}
+    properties:
+      valueType: "Object"
+      readWrite: "R"
   - name: "StreamURI"
     description: "Get video-streaming URI."
     attributes:
--- a/cmd/res/profiles/general.usb.camera.yaml
+++ b/cmd/res/profiles/general.usb.camera.yaml
@@ -44,8 +44,8 @@ deviceResources:
     properties:
       valueType: "Object"
       readWrite: "R"
-  - name: "FramerateFormats"
-    description: "Enumerate supported framerate settings, see https://linuxtv.org/downloads/v4l-dvb-apis/userspace-api/v4l/vidioc-enum-frameintervals.html"
+  - name: "FrameRateFormats"
+    description: "Enumerate supported frame rate settings, see https://linuxtv.org/downloads/v4l-dvb-apis/userspace-api/v4l/vidioc-enum-frameintervals.html"
     attributes:
       { command: "METADATA_FRAMERATE_FORMATS" }
     properties:
@@ -87,15 +87,15 @@ deviceResources:
       valueType: "Bool"
       readWrite: "W"
       defaultValue: "false"
-  - name: "SetFramerate"
-    description: "Set the stream Framerate rate"
+  - name: "SetFrameRate"
+    description: "Set the stream frame rate rate"
     attributes: 
       { command: "VIDEO_SET_FRAMERATE"}
     properties:
       valueType: "Object"
       readWrite: "W"
-  - name: "GetFramerate"
-    description: "Get the stream Framerate rate"
+  - name: "GetFrameRate"
+    description: "Get the stream frame rate rate"
     attributes: 
       { command: "VIDEO_GET_FRAMERATE"}
     properties:
--- a/cmd/res/profiles/general.usb.camera.yaml
+++ b/cmd/res/profiles/general.usb.camera.yaml
@@ -88,14 +88,14 @@ deviceResources:
       readWrite: "W"
       defaultValue: "false"
   - name: "SetFrameRate"
-    description: "Set the stream frame rate rate"
+    description: "Set the stream frame rate"
     attributes: 
       { command: "VIDEO_SET_FRAMERATE"}
     properties:
       valueType: "Object"
       readWrite: "W"
   - name: "GetFrameRate"
-    description: "Get the stream frame rate rate"
+    description: "Get the stream frame rate"
     attributes: 
       { command: "VIDEO_GET_FRAMERATE"}
     properties:
--- a/cmd/res/profiles/general.usb.camera.yaml
+++ b/cmd/res/profiles/general.usb.camera.yaml
@@ -47,7 +47,7 @@ deviceResources:
   - name: "FramerateFormats"
     description: "Enumerate supported framerate settings, see https://linuxtv.org/downloads/v4l-dvb-apis/userspace-api/v4l/vidioc-enum-frameintervals.html"
     attributes:
-      { command: "METADATA_FPS_FORMATS" }
+      { command: "METADATA_FRAMERATE_FORMATS" }
     properties:
       valueType: "Object"
       readWrite: "R"
--- a/cmd/res/profiles/general.usb.camera.yaml
+++ b/cmd/res/profiles/general.usb.camera.yaml
@@ -44,8 +44,8 @@ deviceResources:
     properties:
       valueType: "Object"
       readWrite: "R"
-  - name: "FpsFormats"
-    description: "Enumerate supported fps settings, see https://linuxtv.org/downloads/v4l-dvb-apis/userspace-api/v4l/vidioc-enum-frameintervals.html"
+  - name: "FramerateFormats"
+    description: "Enumerate supported framerate settings, see https://linuxtv.org/downloads/v4l-dvb-apis/userspace-api/v4l/vidioc-enum-frameintervals.html"
     attributes:
       { command: "METADATA_FPS_FORMATS" }
     properties:
@@ -87,8 +87,8 @@ deviceResources:
       valueType: "Bool"
       readWrite: "W"
       defaultValue: "false"
-  - name: "SetFps"
-    description: "Set the stream FPS rate"
+  - name: "SetFramerate"
+    description: "Set the stream Framerate rate"
     attributes: 
       { command: "VIDEO_SET_FPS"}
     properties:
--- a/cmd/res/profiles/general.usb.camera.yaml
+++ b/cmd/res/profiles/general.usb.camera.yaml
@@ -88,7 +88,7 @@ deviceResources:
       readWrite: "W"
       defaultValue: "false"
   - name: "FrameRate"
-    description: "Set the stream frame rate"
+    description: "Set and get the stream frame rate"
     attributes: 
         getFunction: "VIDEO_GET_FRAMERATE"
         setFunction: "VIDEO_SET_FRAMERATE"
--- a/cmd/res/profiles/general.usb.camera.yaml
+++ b/cmd/res/profiles/general.usb.camera.yaml
@@ -12,7 +12,7 @@ deviceResources:
       Camera information including driver name, device name, bus info, and capabilities.
       See https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/vidioc-querycap.html.
     attributes:
-      { command: "METADATA_DEVICE_CAPABILITY" }
+      { getFunction: "METADATA_DEVICE_CAPABILITY" }
     properties:
       valueType: "Object"
       readWrite: "R"
@@ -21,7 +21,7 @@ deviceResources:
       Query or select the current video input.
       See https://linuxtv.org/downloads/v4l-dvb-apis/userspace-api/v4l/vidioc-g-input.html.
     attributes:
-      { command: "METADATA_CURRENT_VIDEO_INPUT" }
+      { getFunction: "METADATA_CURRENT_VIDEO_INPUT" }
     properties:
       valueType: "Int8"
       readWrite: "R"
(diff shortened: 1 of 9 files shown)


vyshali-chitikeshi:
Validation looks good with get frame rate, get video data format and set fps api's.

internal/driver/constants.go
```diff
@@ -9,6 +9,7 @@ package driver
 const (
```
presatish, line 9: Need to update the copyright for this file.

internal/driver/metadata.go
```diff
@@ -135,6 +136,20 @@ func getDataFormat(d *usbdevice.Device) (interface{}, error) {
 		}
 	}
 	result.Quantization = quant
+	intervalCount := 0
```
presatish, line 139: Need to update copyright for this file too.

cmd/res/profiles/general.usb.camera.yaml
```diff
@@ -44,6 +44,13 @@ deviceResources:
     properties:
       valueType: "Object"
       readWrite: "R"
+  - name: "FramerateFormats"
```
presatish: I think `FrameRateFormats` would be more clear and better.

cmd/res/profiles/general.usb.camera.yaml
```diff
@@ -80,6 +87,20 @@ deviceResources:
       valueType: "Bool"
       readWrite: "W"
       defaultValue: "false"
+  - name: "SetFramerate"
+    description: "Set the stream Framerate rate"
```
presatish:
```suggestion
    description: "Set the stream Frame rate"
```

cmd/res/profiles/general.usb.camera.yaml
```diff
@@ -80,6 +87,20 @@ deviceResources:
...
+    attributes: 
+      { command: "VIDEO_SET_FRAMERATE"}
+    properties:
+      valueType: "Object"
+      readWrite: "W"
+  - name: "GetFramerate"
+    description: "Get the stream Framerate rate"
```
presatish:
```suggestion
    description: "Get the stream Frame rate"
```

presatish (changes requested):
Posting first set of my review comments.

cmd/res/profiles/general.usb.camera.yaml
```diff
@@ -44,6 +44,13 @@ deviceResources:
     properties:
       valueType: "Object"
       readWrite: "R"
+  - name: "FrameRateFormats"
+    description: "Enumerate supported frame rate settings, see https://linuxtv.org/downloads/v4l-dvb-apis/userspace-api/v4l/vidioc-enum-frameintervals.html"
+    attributes:
+      { command: "METADATA_FRAMERATE_FORMATS" }
```
ajcasagrande: not a fan of the curly braces here, is there a reason for them?

cmd/res/profiles/general.usb.camera.yaml
```diff
@@ -80,6 +87,20 @@ deviceResources:
...
+    properties:
+      valueType: "Object"
+      readWrite: "W"
+  - name: "GetFrameRate"
+    description: "Get the stream frame rate"
+    attributes: 
+      { command: "VIDEO_GET_FRAMERATE"}
+    properties:
+      valueType: "Object"
+      readWrite: "R"
```
ajcasagrande: I think these can be combined into just one `FrameRate` command that is `RW`. You can look at the onvif profiles to see how it is done there. Something like `getCommand` and `setCommand` in the attributes would work.

docs/USB-Camera-Collection.postman_collection.json
```diff
@@ -251,6 +251,61 @@
...
+						"v3",
+						"device",
+						"name",
+						"{{camera_name}}",
+						"FrameRateFormats"
+					],
+					"query": [
+						{
+							"key": "PathIndex",
+							"value": "0"
```
ajcasagrande: In theory this can be a postman environment variable

internal/driver/device.go
```diff
@@ -89,6 +91,41 @@ func (dev *Device) StopStreaming() {
 	}
 }
 
+// SetFps updates the fps on the device side of the service. Note that this won't update the rtsp output stream fps
+func (dev *Device) SetFps(device *usbdevice.Device, fpsNumerator uint32, fpsDenominator uint32) (string, error) {
+	fps := fmt.Sprintf("%f", float32(fpsDenominator)/float32(fpsNumerator))
+
+	dataFormat, err := getDataFormat(device)
+	if err != nil {
+		return "", nil
```
ajcasagrande: is there a reason this is not returning the error?

internal/driver/device.go
```diff
@@ -89,6 +91,41 @@ func (dev *Device) StopStreaming() {
...
+	}
+	intervals := dataFormat.(DataFormat).FpsIntervals
```
ajcasagrande: this could probably be inlined in the for loop itself.

internal/driver/device.go
```diff
@@ -89,6 +91,41 @@ func (dev *Device) StopStreaming() {
...
+	foundFlag := false
```
ajcasagrande: I would probably just use `found` or `fpsFound`.

internal/driver/device.go
```diff
@@ -89,6 +91,41 @@ func (dev *Device) StopStreaming() {
 	}
 }
 
+// SetFps updates the fps on the device side of the service. Note that this won't update the rtsp output stream fps
+func (dev *Device) SetFps(device *usbdevice.Device, fpsNumerator uint32, fpsDenominator uint32) (string, error) {
```
ajcasagrande: I might change the name of `device` to `usbDevice` to be more specific since there are multiple device objects

internal/driver/driver.go
```diff
@@ -226,20 +226,13 @@ func (d *Driver) HandleReadCommands(deviceName string, protocols map[string]mode
 	reqs []sdkModels.CommandRequest) ([]*sdkModels.CommandValue, error) {
 	d.lc.Debugf("Driver.HandleReadCommands: protocols: %v resource: %v attributes: %v", protocols,
 		reqs[0].DeviceResourceName, reqs[0].Attributes)
-	var err error
+	// var err error
```
ajcasagrande: delete if unused

internal/driver/driver.go
```diff
@@ -286,6 +307,18 @@ func (d *Driver) HandleReadCommands(deviceName string, protocols map[string]mode
...
 			}
 			cv, err = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeObject, data)
+		case MetadataFrameRateFormats:
+			data, err = getSupportedIntervalFormats(cameraDevice)
+			if err != nil {
+				return responses, errorWrapper.CommandError(command, err)
+			}
+			cv, err = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeObject, data)
+		case VideoGetFrameRate:
+			data, err = cameraDevice.GetFrameRate()
```
ajcasagrande: should we make our own function for this since it doesn't support numerators other than 1?

internal/driver/driver.go
```diff
@@ -335,6 +368,40 @@ func (d *Driver) HandleWriteCommands(deviceName string, protocols map[string]mod
...
+				return err
+			}
+			if pathIndexConv >= len(device.paths) {
+				return errors.NewCommonEdgeX(errors.KindIOError,
+					fmt.Sprintf("Video streaming path does not exist for the device %v at PathIndex %d", device.name, pathIndexConv), nil)
+			}
+			videoPath = device.paths[pathIndexConv]
+		}
+
+		// currently defaults to using the first available stream
```
ajcasagrande: comment should probably be moved up to line 384 ish

internal/driver/driver.go
```diff
@@ -351,6 +418,37 @@ func (d *Driver) HandleWriteCommands(deviceName string, protocols map[string]mod
 			}
 		case VideoStopStreaming:
 			device.StopStreaming()
+		case VideoSetFrameRate:
+			fpsParam, edgexErr := params[i].ObjectValue()
+			if edgexErr != nil {
+				return errors.NewCommonEdgeXWrapper(edgexErr)
+			}
+			fpsValueDenominator, ok := fpsParam.(map[string]interface{})[FpsValueDenominator]
```
ajcasagrande: the naming of this in the API input can be a little confusing. technically these are the "interval" values which are an inverse of the actual fps. its actually spf (sunblock anyone)? seconds per frame. I think maybe we name it `FpsInvervalDenominator`? Either way we probably need to document it.

ajcasagrande (changes requested):
Still looking at `driver/metadata.go`. Here are my thoughts so far.

internal/driver/metadata.go
```diff
@@ -100,41 +101,55 @@ func getInputStatus(d *usbdevice.Device, index string) (uint32, error) {
...
+	var intervals []v4l2.Fract
+	for {
+		fd := d.Fd()
+		index := uint32(intervalCount)
+		encoding := pixFmt.PixelFormat
+		if interval, exit := v4l2.GetFormatFrameInterval(fd, index, encoding, pixFmt.Width, pixFmt.Height); exit == nil {
```
ajcasagrande: why is it called `exit` and not `err`?

internal/driver/metadata.go
```diff
@@ -198,3 +213,43 @@ func getImageFormats(d *usbdevice.Device) (interface{}, error) {
...
+		for j, frameSize := range fss {
+			intervalCount := 0
+			var intervals []v4l2.FrameIntervalEnum
+			for {
+				fd := d.Fd()
+				index := uint32(intervalCount)
+				encoding := frameSize.PixelFormat
+				height := frameSize.Size.MaxHeight
+				width := frameSize.Size.MaxWidth
+				if interval, exit := v4l2.GetFormatFrameInterval(fd, index, encoding, width, height); exit == nil {
```
ajcasagrande: why is it called `exit` and not `err`?

internal/driver/metadata.go
```diff
@@ -198,3 +213,43 @@ func getImageFormats(d *usbdevice.Device) (interface{}, error) {
...
 }
+
+func getSupportedIntervalFormats(d *usbdevice.Device) (interface{}, error) {
+	descs, err := d.GetFormatDescriptions()
+	if err != nil {
+		return nil, err
+	}
+	type result struct {
+		IntervalFormats [][][]v4l2.FrameIntervalEnum
+	}
```
ajcasagrande: I am confused on the result format and why the triple nested array? Shouldn't it be more of an array of objects (which have sub arrays)? This way its hard to tell what the distinction between separate arrays means.

cmd/res/profiles/general.usb.camera.yaml
ChristianDarr-personal: I don't know, I just followed the existing formatting. I don't think ONVIF has them

internal/driver/metadata.go
ChristianDarr-personal: I believe that may have been a holdover from a previous iteration where I used a bool.

internal/driver/driver.go
ChristianDarr-personal: Yeah I can take a look at that, shouldn't be too difficult

internal/driver/driver.go
ChristianDarr-personal: Yeah that was a problem during development, the fact that internally it's handled opposite of how we usually think about frame rates.

internal/driver/driver.go
```diff
@@ -226,32 +226,57 @@ func (d *Driver) HandleReadCommands(deviceName string, protocols map[string]mode
...
+			videoPath = device.paths[pathIndexConv]
+		}
+
+		// currently defaults to using the first available stream
+		cameraDevice, err := usbDevice.Open(videoPath)
+		if err != nil {
+			return responses, errors.NewCommonEdgeX(errors.KindServerError,
+				fmt.Sprintf("failed to open the underlying device at specified path %s", videoPath), err)
+		}
+		defer cameraDevice.Close()
```
presatish: I see warning in my IDE for this for a possible resource leak.

internal/driver/driver.go
```diff
@@ -329,12 +361,46 @@ func (d *Driver) HandleWriteCommands(deviceName string, protocols map[string]mod
...
+			}
+			videoPath = device.paths[pathIndexConv]
+		}
+
+		cameraDevice, err := usbDevice.Open(videoPath)
+		if err != nil {
+			return errors.NewCommonEdgeX(errors.KindServerError,
+				fmt.Sprintf("failed to open the underlying device at specified path %s", videoPath), err)
+		}
+		defer cameraDevice.Close()
```
presatish: same for this too as above.

presatish: Also this whole code of reading the request, flushing out the query till opening the device for streaming can be made into a separate function as this is repeated 2 times in the same file.

internal/driver/driver.go
ChristianDarr-personal: Does it give more details? This is the standard practice for opening and closing the device across the service

presatish: `Possible resource leak, 'defer' is called in the 'for' loop`

presatish: I think defer should be removed and it should be just `cameraDevice.Close()` or you probably need to move this part of opening and closing into a separate function.

internal/driver/device.go
```diff
@@ -92,39 +92,50 @@ func (dev *Device) StopStreaming() {
...
-	err = device.SetStreamParam(origStreamParam)
+	err = usbdevice.SetStreamParam(origStreamParam)
 	if err != nil {
 		return "", err
 	}
 
 	return fps, nil
 }
 
+func (dev *Device) GetFrameRate(usbdevice *usbdevice.Device) (v4l2.Fract, error) {
```
ajcasagrande: the name for `usbdevice` is colliding with the package. the argument should probably be called `usbDevice`.

internal/driver/device.go
```diff
@@ -92,39 +92,50 @@ func (dev *Device) StopStreaming() {
 }
 
 // SetFps updates the fps on the device side of the service. Note that this won't update the rtsp output stream fps
-func (dev *Device) SetFps(device *usbdevice.Device, fpsNumerator uint32, fpsDenominator uint32) (string, error) {
+func (dev *Device) SetFrameRate(usbdevice *usbdevice.Device, fpsNumerator uint32, fpsDenominator uint32) (string, error) {
```
ajcasagrande: fps Num/Denom args should be renamed to match the same as what you renamed them in the calling code.

internal/driver/metadata.go
ajcasagrande: i think this one still applies

internal/driver/metadata.go
```diff
@@ -219,37 +219,53 @@ func getSupportedIntervalFormats(d *usbdevice.Device) (interface{}, error) {
...
+		FrameType   uint32
+		PixelFormat uint32
+		Height      uint32
+		Width       uint32
+		Interval    []v4l2.Fract
+	}
+	type frameFormat struct {
+		Description string
+		Intervals   []intervalsInfo
+	}
```
ajcasagrande: the types should probably be moved outside of the function body.

ajcasagrande (changes requested):

internal/driver/device.go
```diff
@@ -89,6 +91,51 @@ func (dev *Device) StopStreaming() {
 	}
 }
 
+// SetFps updates the fps on the device side of the service. Note that this won't update the rtsp output stream fps
```
ajcasagrande: comment needs to be updated with correct function name

internal/driver/driver.go
ajcasagrande: yes, what will happen is the defer is not called until the entire for loop is completed, where what you really want is it to be closed at the end of each for iteration.

this can be addressed by either not using defer and closing it at the end (if applicable), or moving part of the body of the for loop into a separate function that can have a defer inside of it.

internal/driver/metadata.go
```diff
@@ -214,58 +217,50 @@ func getImageFormats(d *usbdevice.Device) (interface{}, error) {
...
+			frameInfo.Width = width
+			frameInfo.PixelFormat = encoding
+			frameInfo.Index = frameSize.Index
 			for {
 				fd := d.Fd()
 				index := uint32(intervalCount)
-				if interval, exit := v4l2.GetFormatFrameInterval(fd, index, encoding, width, height); exit == nil {
-					frameIntervals.Interval = append(frameIntervals.Interval, interval.Interval.Max)
+				if interval, err := v4l2.GetFormatFrameInterval(fd, index, encoding, width, height); err == nil {
+					frameInfo.Rates = append(frameInfo.Rates, v4l2.Fract{
```
ajcasagrande, line 254: I would add a small comment in the couple of places that you do the swapping of num/denom with denom/num such that others looking at the code do not think it is a typo.

internal/driver/metadata.go
```diff
@@ -137,19 +137,22 @@ func getDataFormat(d *usbdevice.Device) (interface{}, error) {
...
-	var intervals []v4l2.Fract
+	var frameRates []v4l2.Fract
 	for {
 		fd := d.Fd()
 		index := uint32(intervalCount)
 		encoding := pixFmt.PixelFormat
 		if interval, err := v4l2.GetFormatFrameInterval(fd, index, encoding, pixFmt.Width, pixFmt.Height); err == nil {
 			intervalCount += 1
-			intervals = append(intervals, interval.Interval.Max)
+			frameRates = append(frameRates, v4l2.Fract{
```
ajcasagrande, line 149: I would add a small comment in the couple of places that you do the swapping of num/denom with denom/num such that others looking at the code do not think it is a typo.

internal/driver/device.go
```diff
@@ -91,16 +91,16 @@ func (dev *Device) StopStreaming() {
 	}
 }
 
-// SetFps updates the fps on the device side of the service. Note that this won't update the rtsp output stream fps
-func (dev *Device) SetFrameRate(usbdevice *usbdevice.Device, fpsNumerator uint32, fpsDenominator uint32) (string, error) {
-	fps := fmt.Sprintf("%f", float32(fpsDenominator)/float32(fpsNumerator))
-	dataFormat, err := getDataFormat(usbdevice)
+// SetFrameRate updates the fps on the device side of the service. Note that this won't update the rtsp output stream fps
+func (dev *Device) SetFrameRate(usbDevice *usbdevice.Device, intervalNumerator uint32, intervalDenominator uint32) (string, error) {
```
ajcasagrande: i would switch this function to just taking the fps values as is, and comparing them as is, and then just invert when setting the timePerFrame.

internal/driver/driver.go
```diff
@@ -244,114 +241,112 @@ func (d *Driver) HandleReadCommands(deviceName string, protocols map[string]mode
...
+}
 
-		// currently defaults to using the first available stream if the user does not provide input
-		cameraDevice, err := usbDevice.Open(videoPath)
-		if err != nil {
-			return responses, errors.NewCommonEdgeX(errors.KindServerError,
-				fmt.Sprintf("failed to open the underlying device for streaming at specified path %s", videoPath), err)
-		}
-		defer cameraDevice.Close()
+func (d *Driver) SelectReadCommands(device *Device, req sdkModels.CommandRequest, command interface{}) (*sdkModels.CommandValue, error) {
```
ajcasagrande: should the name be more like `ExecuteReadCommand`?

internal/driver/driver.go
```diff
@@ -368,92 +363,88 @@ func (d *Driver) HandleWriteCommands(deviceName string, protocols map[string]mod
...
-				return errors.NewCommonEdgeX(errors.KindIOError,
-					fmt.Sprintf("Video streaming path does not exist for the device %v at PathIndex %d", device.name, pathIndexConv), nil)
-			}
-			videoPath = device.paths[pathIndexConv]
-		}
+	}
+	return nil
+}
+
+func (d *Driver) SelectWriteCommands(device *Device, req sdkModels.CommandRequest, i int, params []*sdkModels.CommandValue, command interface{}) error {
```
ajcasagrande: should the name be more like `ExecuteWriteCommand`?

ajcasagrande (changes requested):

vyshali-chitikeshi:
Validation looks good

presatish (approved):
LGTM

ajcasagrande (approved):

PR #266, opened 2023-08-03
feat: select stream for api commands

ChristianDarr-personal:

## PR Checklist

- [ ] I have added unit tests for the new feature or bug fix (if not, why?) I've only added device level functionality
  <link to docs PR>

## Testing Instructions

1. Execute each command with no query parameters and see status code 200.
2. Execute each command with the `PathIndex` set to the appropriate value. For standard usb cameras, this can only be 0, and for RealSense cameras this can be 0, 1, or 2. If the index is outside of that range, it will send an error.
3. Execute each command with the `StreamFormat` set to the appropriate value. For standard usb cameras, this can only be "RGB", and for RealSense cameras this can be "RGB", "Greyscale" or "Depth". If the value is not supported, it will send an error.
4. For realsense, start streaming for each index and check that the stream is working (using vlc or mplayer) and there are no errors. Also make sure that the streams are accurate, ie rgb shows rgb, etc. Make sure to stop streaming after checking the stream.
Note: Some commands will return the "/dev/video" stream. Make sure that matches the stream that you supplied in the query parameter.
Note: The stop streaming command does not need one of these query parameters.

### DIFF
--- a/docs/USB-Camera-Collection.postman_collection.json
+++ b/docs/USB-Camera-Collection.postman_collection.json
@@ -821,7 +821,13 @@
 					"query": [
 						{
 							"key": "PathIndex",
-							"value": "{{path_index}}"
+							"value": "{{path_index}}",
+							"disabled": true
+						},
+						{
+							"key": "StreamFormat",
+							"value": "{{stream_format}}",
+							"disabled": true
 						}
 					]
 				}
@@ -873,7 +879,13 @@
 					"query": [
 						{
 							"key": "PathIndex",
-							"value": "{{path_index}}"
+							"value": "{{path_index}}",
+							"disabled": true
+						},
+						{
+							"key": "StreamFormat",
+							"value": "{{stream_format}}",
+							"disabled": true
 						}
 					]
 				},
--- a/docs/USB-Camera-Collection.postman_collection.json
+++ b/docs/USB-Camera-Collection.postman_collection.json
@@ -1030,7 +1030,7 @@
 					}
 				},
 				"url": {
-					"raw": "http://{{command_host}}:{{command_port}}/api/v3/device/name/{{camera_name}}/CameraStatus",
+					"raw": "http://{{command_host}}:{{command_port}}/api/v3/device/name/{{camera_name}}/CameraStatus?InputIndex=0",
 					"protocol": "http",
 					"host": [
 						"{{command_host}}"
@@ -1046,7 +1046,11 @@
 					],
 					"query": [
 						{
-							"key": "InputIndex",
+                            "key": "InputIndex",
+                            "value": "0"
+						},
+                        {
+							"key": "PathIndex",
 							"value": "{{path_index}}",
 							"disabled": true
 						},
--- a/docs/USB-Camera-Collection.postman_collection.json
+++ b/docs/USB-Camera-Collection.postman_collection.json
@@ -636,18 +636,6 @@
 						"name",
 						"{{camera_name}}",
 						"StreamingStatus"
-					],
-					"query": [
-						{
-							"key": "PathIndex",
-							"value": "{{path_index}}",
-							"disabled": true
-						},
-						{
-							"key": "StreamFormat",
-							"value": "{{stream_format}}",
-							"disabled": true
-						}
 					]
 				}
 			},
--- a/docs/USB-Camera-Collection.postman_collection.json
+++ b/docs/USB-Camera-Collection.postman_collection.json
@@ -99,18 +99,6 @@
 						"name",
 						"{{camera_name}}",
 						"CameraInfo"
-					],
-					"query": [
-						{
-							"key": "PathIndex",
-							"value": "{{path_index}}",
-							"disabled": true
-						},
-						{
-							"key": "StreamFormat",
-							"value": "{{stream_format}}",
-							"disabled": true
-						}
 					]
 				}
 			},
--- a/docs/USB-Camera-Collection.postman_collection.json
+++ b/docs/USB-Camera-Collection.postman_collection.json
@@ -993,18 +993,6 @@
 						"name",
 						"{{camera_name}}",
 						"StopStreaming"
-					],
-					"query": [
-						{
-							"key": "PathIndex",
-							"value": "{{path_index}}",
-							"disabled": true
-						},
-						{
-							"key": "StreamFormat",
-							"value": "{{stream_format}}",
-							"disabled": true
-						}
 					]
 				},
 				"description": "Test steps:\n\n1) Execute start streaming api with valid input and outputs\n2) Execute get stream status api and verify that \"IsStreaming\" set to true\n3) Execute stop stream api\n4) Execute get stream status api and verify that \"IsStreaming\" set to false\nNote: Camera shows green light in front when streaming started"
--- a/docs/USB-Camera-Collection.postman_collection.json
+++ b/docs/USB-Camera-Collection.postman_collection.json
@@ -1,6 +1,6 @@
 {
 	"info": {
-		"_postman_id": "1cd1947d-95fa-456b-ae95-54b38b90723d",
+		"_postman_id": "12557ee1-fc0d-4e65-80f6-22633bc34747",
 		"name": "USB-Camera-Collection",
 		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
 	},
@@ -99,6 +99,18 @@
 						"name",
 						"{{camera_name}}",
 						"CameraInfo"
+					],
+					"query": [
+						{
+							"key": "PathIndex",
+							"value": "{{path_index}}",
+							"disabled": true
+						},
+						{
+							"key": "StreamFormat",
+							"value": "{{stream_format}}",
+							"disabled": true
+						}
 					]
 				}
 			},
@@ -148,6 +160,18 @@
 						"name",
 						"{{camera_name}}",
 						"VideoInputIndex"
+					],
+					"query": [
+						{
+							"key": "PathIndex",
+							"value": "{{path_index}}",
+							"disabled": true
+						},
+						{
+							"key": "StreamFormat",
+							"value": "{{stream_format}}",
+							"disabled": true
+						}
 					]
 				}
 			},
@@ -197,6 +221,18 @@
 						"name",
 						"{{camera_name}}",
 						"ImageFormats"
+					],
+					"query": [
+						{
+							"key": "PathIndex",
+							"value": "{{path_index}}",
+							"disabled": true
+						},
+						{
+							"key": "StreamFormat",
+							"value": "{{stream_format}}",
+							"disabled": true
+						}
 					]
 				}
 			},
@@ -233,7 +269,7 @@
 					}
 				},
 				"url": {
-					"raw": "http://{{command_host}}:{{command_port}}/api/v3/device/name/{{camera_name}}/DataFormat?PathIndex={{path_index}}",
+					"raw": "http://{{command_host}}:{{command_port}}/api/v3/device/name/{{camera_name}}/DataFormat",
 					"protocol": "http",
 					"host": [
 						"{{command_host}}"
@@ -250,7 +286,13 @@
 					"query": [
 						{
 							"key": "PathIndex",
-							"value": "{{path_index}}"
+							"value": "{{path_index}}",
+							"disabled": true
+						},
+						{
+							"key": "StreamFormat",
+							"value": "{{stream_format}}",
+							"disabled": true
 						}
 					]
 				}
@@ -288,7 +330,7 @@
 					}
 				},
 				"url": {
-					"raw": "http://{{command_host}}:{{command_port}}/api/v3/device/name/{{camera_name}}/FrameRateFormats?PathIndex={{path_index}}",
+					"raw": "http://{{command_host}}:{{command_port}}/api/v3/device/name/{{camera_name}}/FrameRateFormats",
 					"protocol": "http",
 					"host": [
 						"{{command_host}}"
@@ -305,7 +347,13 @@
 					"query": [
 						{
 							"key": "PathIndex",
-							"value": "{{path_index}}"
+							"value": "{{path_index}}",
+							"disabled": true
+						},
+						{
+							"key": "StreamFormat",
+							"value": "{{stream_format}}",
+							"disabled": true
 						}
 					]
 				}
@@ -356,6 +404,18 @@
 						"name",
 						"{{camera_name}}",
 						"CropCapability"
+					],
+					"query": [
+						{
+							"key": "PathIndex",
+							"value": "{{path_index}}",
+							"disabled": true
+						},
+						{
+							"key": "StreamFormat",
+							"value": "{{stream_format}}",
+							"disabled": true
+						}
 					]
 				}
 			},
@@ -405,6 +465,18 @@
 						"name",
 						"{{camera_name}}",
 						"StreamingParam"
+					],
+					"query": [
+						{
+							"key": "PathIndex",
+							"value": "{{path_index}}",
+							"disabled": true
+						},
+						{
+							"key": "StreamFormat",
+							"value": "{{stream_format}}",
+							"disabled": true
+						}
 					]
 				}
 			},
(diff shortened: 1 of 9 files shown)


internal/driver/driver.go
```diff
@@ -805,6 +872,24 @@ func (d *Driver) newDevice(name string, protocols map[string]models.ProtocolProp
...
+		return errors.NewCommonEdgeX(errors.KindServerError,
+			fmt.Sprintf("failed to initialize new transcoder for device %s", device.name), err)
+	}
+	// // Initialize transcoder passing the input path and output path, along with the credentials
+	// err = trans.Initialize(fdPath, d.getAuthenticatedRTSPUri(name))
+	// if err != nil {
+	// 	return errors.NewCommonEdgeX(errors.KindServerError,
+	// 		fmt.Sprintf("failed to initialize new transcoder for device %s", name), err)
+	// }
+	// trans.MediaFile().SetOutputFormat(RtspUriScheme)
```
marcpfuller: Do we need this commented code?

internal/driver/driver.go
```diff
@@ -491,10 +507,61 @@ func (d *Driver) getPathName(device *Device, queryParams url.Values) (string, er
 				fmt.Sprintf("Video streaming path does not exist for the device %v at PathIndex %d", device.name, pathIndexConv), nil)
 		}
 		videoPath = device.paths[pathIndexConv]
+	} else if len(streamFormat) != 0 { // use stream format video path value
+		if streamFormat == RGB || streamFormat == Greyscale || streamFormat == Depth {
```
ajcasagrande: I would invert this if and return the error, and then you do not have to nest as much

```go
if streamFormat != RGB && streamFormat != Greyscale && streamFormat != Depth {
	return error...
}

for ... {
}
```

vyshali-chitikeshi:
Everything looks good from validation perspective. Below are details
Imported postman collection from this PR branch and ran below tests.
Standard USB camera:
Executed each API command in collection without any query parameters - Pass
Executed each API command in collection with valid pathindex query- Pass
Executed some of API commands (start streaming and some other API's randomly tested) in collection with invalid path index (negative test)- Pass
Executed each API command in collection with valid stream format query- Pass
Executed some of API commands  (start streaming and some other API's randomly tested) in collection with invalid stream format query(negative test)- Pass
Real sense camera:
Standard USB camera:
1.  Executed each API command in collection without any query parameters - Pass
2. Executed each API command in collection with valid path index query(0,1,2)- Pass
3. Executed some of API commands (start streaming and some other API's randomly tested) in collection with invalid path index (negative test)- Pass
4. Executed each API command in collection with valid stream format query("RGB", Greyscale","Depth")- Pass
5. Executed some of API commands  (start streaming and some other API's randomly tested) in collection with invalid stream format query(negative test)- Pass
6. Checking its executing stream matches with  input query parameters given- Pass

internal/driver/driver.go
```diff
@@ -266,7 +273,7 @@ func (d *Driver) ExecuteReadCommands(device *Device, req sdkModels.CommandReques
 
 	videoPath, err := d.getPathName(device, queryParams)
 	if err != nil {
-		return nil, err
+		return cv, err
```
ajcasagrande: this should probably stay as nil

internal/driver/driver.go
```diff
@@ -478,10 +492,12 @@ func (d *Driver) Stop(force bool) error {
 func (d *Driver) getPathName(device *Device, queryParams url.Values) (string, error) {
 	var videoPath string
 	pathIndex := queryParams.Get(PathIndex)
-	if len(pathIndex) == 0 {
-		// currently defaults to using the first available stream
+	streamFormat := queryParams.Get(StreamFormat)
+	if len(pathIndex) == 0 && len(streamFormat) == 0 { // most common case
```
ajcasagrande: if these are strings, just use `if pathIndex == ""`, etc.

this is used a lot it looks like

internal/driver/driver.go
```diff
@@ -491,10 +507,61 @@ func (d *Driver) getPathName(device *Device, queryParams url.Values) (string, er
...
 		videoPath = device.paths[pathIndexConv]
+	} else if len(streamFormat) != 0 { // use stream format video path value
+		if streamFormat == RGB || streamFormat == Greyscale || streamFormat == Depth {
+			for _, path := range device.paths {
+				videoPath, err := getStreamFormatPath(path, streamFormat)
+				if err != nil {
+					continue
+				} else {
+					return videoPath, nil
+				}
```
ajcasagrande: dont need it to be in an else block

internal/driver/driver.go
```diff
@@ -805,6 +872,24 @@ func (d *Driver) newDevice(name string, protocols map[string]models.ProtocolProp
 	}, nil
 }
 
+func (d *Driver) updateTranscoder(device *Device, fdPath string) error {
```
ajcasagrande: this function needs cleaning up

internal/driver/metadata.go
```diff
@@ -105,27 +107,28 @@ func getDataFormat(d *usbdevice.Device) (interface{}, error) {
...
+	pixDescription, _ := getCompleteFormatMapEntry(pixFmt.PixelFormat)
+	dataFormat := DataFormat{}
+	dataFormat.Height = pixFmt.Height
+	dataFormat.Width = pixFmt.Width
+	dataFormat.PixelFormat = pixFmt.PixelFormat
+	dataFormat.PixelFormatDescription = pixDescription
+	dataFormat.Field = v4l2.Fields[pixFmt.Field]
+	dataFormat.BytesPerLine = pixFmt.BytesPerLine
+	dataFormat.SizeImage = pixFmt.SizeImage
+	dataFormat.Colorspace = v4l2.Colorspaces[pixFmt.Colorspace]
```
ajcasagrande: use object initializer:

```go
DataFormat {
 Height: pixFmt.Height,
 ...
}
```

internal/driver/metadata.go
```diff
@@ -173,22 +179,25 @@ func getStreamingParameters(d *usbdevice.Device) (interface{}, error) {
...
-	result.CaptureMode = CaptureMode{hiqual, sp.Capture.CaptureMode}
+	streamingParameters.CaptureMode = CaptureMode{hiqual, sp.Capture.CaptureMode}
+
+	streamingParameters.TimePerFrame = sp.Capture.TimePerFrame
+	streamingParameters.ReadBuffers = sp.Capture.ReadBuffers
 
-	result.TimePerFrame = sp.Capture.TimePerFrame
-	result.ReadBuffers = sp.Capture.ReadBuffers
+	result := make(map[string]StreamingParameters)
+	result[d.Name()] = streamingParameters
```
ajcasagrande, line 200: what is this doing? formatting it as the API would return it?

internal/driver/metadata.go
```diff
@@ -266,5 +279,37 @@ func getSupportedFrameRateFormats(d *usbdevice.Device) (interface{}, error) {
...
+	return result, nil
+}
+
+func getCompleteFormatMapEntry(key uint32) (string, bool) {
+	unsupported := map[uint32]string{
+		PixelFmtDepth:  PixelFmtDepthDesc,
+		PixelFmtUYVY:   PixelFmtUYVYDesc,
+		PixelFmtGrey8:  PixelFmtGrey8Desc,
+		PixelFmtGrey12: PixelFmtGrey12Desc,
+	}
```
ajcasagrande: this could be in a var block at the top of the file, instead of defined every time in the function:

```go
var (
	unsupportedPixelFmts = map[uint32]string{
		PixelFmtDepth:  PixelFmtDepthDesc,
		PixelFmtUYVY:   PixelFmtUYVYDesc,
		PixelFmtGrey8:  PixelFmtGrey8Desc,
		PixelFmtGrey12: PixelFmtGrey12Desc,
	}

)

File: internal/driver/metadata.go [hunk 10]
```diff
@@ -266,5 +279,37 @@ func getSupportedFrameRateFormats(d *usbdevice.Device) (interface{}, error) {
...
+	streamParam, err := d.GetStreamParam()
+	if err != nil {
+		return v4l2.Fract{}, err
+	}
+	timePerFrame := streamParam.Capture.TimePerFrame
+	var fps v4l2.Fract
+	fps.Denominator = timePerFrame.Numerator
+	fps.Numerator = timePerFrame.Denominator
+	result := make(map[string]v4l2.Fract)
+	result[d.Name()] = fps
```
ajcasagrande, line 297: these lines are used a lot, is there a way to refactor these out to the caller? also I think it would make sense to change the return type to be specific if possible.

internal/driver/metadata.go
```diff
@@ -266,5 +279,37 @@ func getSupportedFrameRateFormats(d *usbdevice.Device) (interface{}, error) {
...
-	return r, nil
+	resultMap := make(map[string]result)
+	resultMap[d.Name()] = r
+	return resultMap, nil
+}
+
+func GetFrameRate(d *usbdevice.Device) (interface{}, error) {
+	streamParam, err := d.GetStreamParam()
+	if err != nil {
+		return v4l2.Fract{}, err
```
ajcasagrande: should probably be `nil, err` if possible. the return type here does not match below

ajcasagrande (changes requested):

internal/driver/metadata.go
ChristianDarr-personal: It's adding the path name to the data structure so that when the value is returned, its made clear that the information is specifically for that video path
so it looks like this: 
```
 "objectValue": {
      "/dev/video6": {
          "ImageFormats": [
              {
                  "BufType": 1,
                  "Description": "YUYV 4:2:2",
                  "Flags": 0,
                  "FrameSizes": [
                      {
                          "Index": 0,
                          "PixelFormat": 1448695129,
                          "Size": {
                              "MaxHeight": 
```

internal/driver/constants.go
```diff
@@ -76,4 +80,17 @@ const (
 	UdevSerialShort = "ID_SERIAL_SHORT"
 	UdevSerial      = "ID_SERIAL"
 	UdevV4lProduct  = "ID_V4L_PRODUCT"
+
+	// unsupported pixel formats
+	PixelFmtDepth     = 540422490
+	PixelFmtDepthDesc = "16-bit depth"
```
presatish: I do not think you need to hard code these descriptions. We can get these descriptions using `GetFormatDescriptions()` which is what I have done for my story. You can look into my PR.

internal/driver/constants.go
```diff
@@ -76,4 +80,17 @@ const (
...
+
+	PixelFmtUYVY     = 1498831189
+	PixelFmtUYVYDesc = "UYVY 4:2:2"
```
presatish: You do not need this block. V4l2 library supports this. You can look into my PR if needed for reference.

presatish (changes requested):
I have posted some comments, need to address them.

internal/driver/metadata.go
presatish: Along with this there is also `Path` field in the output? Are these both not the same? Also do we really need to format like this, if so then should the same be done for other command outputs?

ChristianDarr-personal: It is done for all of the applicable outputs. Not every metadata command changes based on path.

internal/driver/metadata.go
ChristianDarr-personal: It doesn't look like we can extract it out to the caller, because the command value requires the data needs to be an interface

docs/USB-Camera-Collection.postman_collection.json
```diff
@@ -99,6 +99,18 @@
 						"name",
 						"{{camera_name}}",
 						"CameraInfo"
```
presatish: Why is `pathIndex` and `StreamFormat` required for this command? I think its the same output for all?

internal/driver/driver.go
```diff
@@ -507,20 +506,19 @@ func (d *Driver) getPathName(device *Device, queryParams url.Values) (string, er
...
+		if streamFormat != RGB && streamFormat != Greyscale && streamFormat != Depth {
 			return "", errors.NewCommonEdgeX(errors.KindIOError, "Invalid stream format. Valid options are 'RGB', 'Greyscale', or 'Depth.'", nil)
 		}
+		for _, path := range device.paths {
+			videoPath, err := getStreamFormatPath(path, streamFormat)
+			if err != nil {
+				continue
+			} else {
+				return videoPath, nil
+			}
```
ajcasagrande: this doesn't need to be in an else block

internal/driver/metadata.go
```diff
@@ -266,5 +280,34 @@ func getSupportedFrameRateFormats(d *usbdevice.Device) (interface{}, error) {
...
+	timePerFrame := streamParam.Capture.TimePerFrame
+	var fps v4l2.Fract
+	fps.Denominator = timePerFrame.Numerator
+	fps.Numerator = timePerFrame.Denominator
+	result := make(map[string]v4l2.Fract)
+	result[d.Name()] = fps
+	return result, nil
+}
+
+func getCompleteFormatMapEntry(usbDevice *usbdevice.Device, pixFmt uint32) string {
```
presatish: I think its always good to capture and return even the error.

presatish: The name can be changed into something more clear, like `getFormatDesc`, `getFormatDescription` or `getPixFormatDesc`?

presatish (changes requested):
Posting last set of comments.

docs/USB-Camera-Collection.postman_collection.json
ChristianDarr-personal: Good catch

docs/USB-Camera-Collection.postman_collection.json
```diff
@@ -915,7 +1047,13 @@
 					"query": [
 						{
 							"key": "InputIndex",
-							"value": "0"
+							"value": "{{path_index}}",
```
ajcasagrande, line 1037: The key on this one says `InputIndex`. Should it be `PathIndex`?

internal/driver/driver.go
```diff
@@ -416,20 +440,24 @@ func (d *Driver) HandleWriteCommands(deviceName string, protocols map[string]mod
...
 		if err != nil {
+			// flush query parameter for remaining reqs
+			for _, req := range reqs {
+				if _, ok := req.Attributes[UrlRawQuery]; ok {
+					req.Attributes[UrlRawQuery] = ""
+				}
+			}
 			return err
 		}
-		// flush query parameter
```
ajcasagrande, line 423: should this comment still be here?

internal/driver/driver.go
```diff
@@ -562,10 +596,40 @@ func (d *Driver) getPathName(device *Device, queryParams url.Values) (string, er
...
+				continue
+			}
+			return videoPath, nil
+		}
+		return "", errors.NewCommonEdgeX(errors.KindIOError, fmt.Sprintf("Invalid stream format for device %s.", device.name), nil)
 	}
 	return videoPath, nil
 }
 
+func getStreamFormatPath(path string, streamFormat string) (string, error) {
```
ajcasagrande: 1. Path does not need to be returned, as you are just returning the input argument
2. The name of the function should be more clear, as this is more of a comparison/match than a get
3. I would recommend changing it so that the function returns just a bool whether or not it matches
4. I would make it be a driver receiver to be able to log the errors


```go
func (d *Driver) pathMatchesStreamFormat(path, streamFormat string) bool {
    ...
    if err != nil {
      d.lc.Errorf("blah")
      return false 
   }
...
return  currentType == streamFormat

}
```

And then up above inside of `getPathName`, do something like so:
```go
		for _, path := range device.paths {
			if d.pathMatchesStreamFormat(path, streamFormat) {
                            return path
                        }
		}
```

internal/driver/driver.go
```diff
@@ -878,6 +942,16 @@ func (d *Driver) newDevice(name string, protocols map[string]models.ProtocolProp
 	}, nil
 }
 
+func (d *Driver) updateTranscoder(device *Device, fdPath string) error {
```
ajcasagrande: maybe rename to `updateTranscoderInputPath`. Also, would this make more sense as something inside of the `*Device` struct itself?

ajcasagrande (changes requested):

internal/driver/driver.go
ChristianDarr-personal: Do you mean add the path as a entry in the struct or to make this a device method?

ajcasagrande (dismissed):
LGTM

vyshali-chitikeshi:
Validation looks good with latest code

presatish (approved):

PR #268, opened 2023-08-08
feat: ability to disable RTSP server on startup via configuration

marcpfuller:

## PR Checklist

- [ ] I have added unit tests for the new feature or bug fix (if not, why?)N/A
N/A

## Testing Instructions
1. run `make test`
2.  start up edgex using edgex-compose `make run no-secty`
3. build device-usb-camera `make build`
4. enable/ disable  EnableRtspServer in canfiguration.yml
5. verify Rtsp server starts and does not start when enabled/disabled

### DIFF
--- a/cmd/res/configuration.yaml
+++ b/cmd/res/configuration.yaml
@@ -38,6 +38,7 @@ Device:
     Interval: "1h"
 
 Driver:
+  EnableRtspServer: "true"
   RtspServerHostName: "localhost"
   RtspTcpPort: "8554"
   RtspAuthenticationServer: "localhost:8000"
--- a/internal/driver/constants.go
+++ b/internal/driver/constants.go
@@ -16,6 +16,8 @@ const (
 	AutoStreaming                   = "AutoStreaming"
 	InputIndex                      = "InputIndex"
 	UrlRawQuery                     = "urlRawQuery"
+	EnableRtspServer                = "EnableRtspServer"
+	RtspServerCmd                   = "./rtsp-simple-server"
 	RtspServerHostName              = "RtspServerHostName"
 	DefaultRtspServerHostName       = "localhost"
 	RtspTcpPort                     = "RtspTcpPort"
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -106,7 +106,7 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
 		_, err := os.Stat(RtspServerCmd)
 		if err != nil {
 			if os.IsNotExist(err) {
-				return fmt.Errorf("%s file canont be found: %s", RtspServerCmd, err.Error())
+				return fmt.Errorf("%s file cannot be found: %s", RtspServerCmd, err.Error())
 			}
 		}
 		rtspProc := exec.Command(RtspServerCmd)
--- a/internal/driver/driver.go
+++ b/internal/driver/driver.go
@@ -306,71 +306,79 @@ func (d *Driver) ExecuteReadCommands(device *Device, req sdkModels.CommandReques
 	case MetadataDeviceCapability:
 		data, err = getCapability(cameraDevice)
 		if err != nil {
-			return cv, errorWrapper.CommandError(command, err)
+			return nil, errorWrapper.CommandError(command, err)
 		}
 		cv, err = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeObject, data)
 	case MetadataCurrentVideoInput:
 		data, err = cameraDevice.GetVideoInputIndex()
 		if err != nil {
-			return cv, errorWrapper.CommandError(command, err)
+			return nil, errorWrapper.CommandError(command, err)
 		}
 		cv, err = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeInt32, data)
 	case MetadataCameraStatus:
 		index := queryParams.Get(InputIndex)
 		if len(index) == 0 {
-			return cv, fmt.Errorf("mandatory query parameter %s not found", InputIndex)
+			return nil, fmt.Errorf("mandatory query parameter %s not found", InputIndex)
 		}
 		data, err = getInputStatus(cameraDevice, index)
 		if err != nil {
-			return cv, errorWrapper.CommandError(command, err)
+			return nil, errorWrapper.CommandError(command, err)
 		}
 		cv, err = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeUint32, data)
 	case MetadataImageFormats:
 		data, err = getImageFormats(cameraDevice)
 		if err != nil {
-			return cv, errorWrapper.CommandError(command, err)
+			return nil, errorWrapper.CommandError(command, err)
 		}
 		cv, err = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeObject, data)
 	case MetadataFrameRateFormats:
 		data, err = getSupportedFrameRateFormats(cameraDevice)
 		if err != nil {
-			return cv, errorWrapper.CommandError(command, err)
+			return nil, errorWrapper.CommandError(command, err)
 		}
 		cv, err = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeObject, data)
 	case VideoGetFrameRate:
 		data, err = device.GetFrameRate(cameraDevice)
 		if err != nil {
-			return cv, errorWrapper.CommandError(command, err)
+			return nil, errorWrapper.CommandError(command, err)
 		}
 		cv, err = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeObject, data)
 	case MetadataDataFormat:
 		data, err = getDataFormat(cameraDevice)
 		if err != nil {
-			return cv, errorWrapper.CommandError(command, err)
+			return nil, errorWrapper.CommandError(command, err)
 		}
 		cv, err = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeObject, data)
 	case MetadataCroppingAbility:
 		data, err = getCropInfo(cameraDevice)
 		if err != nil {
-			return cv, errorWrapper.CommandError(command, err)
+			return nil, errorWrapper.CommandError(command, err)
 		}
 		cv, err = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeObject, data)
 	case MetadataStreamingParameters:
 		data, err = getStreamingParameters(cameraDevice)
 		if err != nil {
-			return cv, errorWrapper.CommandError(command, err)
+			return nil, errorWrapper.CommandError(command, err)
 		}
 		cv, err = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeObject, data)
 	case VideoStreamUri:
+		if !d.enableRtspServer {
+			return nil, errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf(
+				"rtsp server is not enabled, cannot get stream URI for device %s", device.name), nil)
+		}
 		cv, err = sdkModels.NewCommandValue(req.DeviceResourceName, req.Type, device.rtspUri)
 
 	case VideoStreamingStatus:
+		if !d.enableRtspServer {
+			return nil, errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf(
+				"rtsp server is not enabled, cannot get streaming status for device %s", device.name), nil)
+		}
 		cv, err = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeObject, device.streamingStatus)
 	default:
-		return cv, errors.NewCommonEdgeX(errors.KindContractInvalid, fmt.Sprintf("unsupported command %s", command), nil)
+		return nil, errors.NewCommonEdgeX(errors.KindContractInvalid, fmt.Sprintf("unsupported command %s", command), nil)
 	}
 	if err != nil {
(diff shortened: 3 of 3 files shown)


vyshali-chitikeshi:
Validation looks good both in secure  and non-secure mode deployments. Below are more details
 1. When rtsp server disabled in configuration.yml file, streaming status, start streaming and stop     
     streaming APIs gives error "Device resource rtsp server is not enabled" as expected.
 2. When rtsp server is enabled in configuration file, streaming status, start and stop streaming gives
     no error and works as expected. Also noticed below msg in logs
     "Starting rtsp authentication server on localhost:8000", This msg does not show up in logs when
      rtsp is disabled.
3. Verified some general api's like camera status, camera info and camera status etc in both cases i.e.
    rtsp server enabled/disabled and they work correctly
4. Negative test: Provided invalid value (something other than true/false) for rtsp server in configuration.yml file , noticed that it gave below error in logs, boot strapping exits before discovering cameras.  I think this is expected.
level=ERROR ts=2023-08-10T19:15:40.54388004Z app=device-usb-camera source=init.go:77 msg="ProtocolDriver init failed: failed to parse EnableRtspServer config: strconv.ParseBool: parsing \"abcd\": invalid syntax"
5. Validated test cases from 1 to 4 both in secure and non secure modes, and results are same.

internal/driver/driver.go
```diff
@@ -94,32 +96,52 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
...
+		return fmt.Errorf("failed to parse EnableRtspServer config: %s", err.Error())
 	}
-	d.lc.Infof("RTSP server hostname: %s", rtspServerHostName)
-	d.rtspHostName = rtspServerHostName
+	if d.enableRtspServer {
+		// check to see if rtsp-simple-server file/binary exists
+		_, err := os.Stat(RtspServerCmd)
+		if err != nil {
+			if os.IsNotExist(err) {
+				return fmt.Errorf("%s file canont be found: %s", RtspServerCmd, err.Error())
```
seanohair22:
```suggestion
				return fmt.Errorf("%s file cannot be found: %s", RtspServerCmd, err.Error())
```

seanohair22 (changes requested):
Tested, looks good! Just found one spelling error...

ajcasagrande:
PR name should be updated. this is a `feat`. Also the current functionality is that it is enabled, so something like this:
```
feat: ability to disable RTSP server on startup via configuration
```

internal/driver/driver.go
```diff
@@ -94,32 +96,52 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
...
 	}
-	d.lc.Infof("RTSP server hostname: %s", rtspServerHostName)
-	d.rtspHostName = rtspServerHostName
+	if d.enableRtspServer {
+		// check to see if rtsp-simple-server file/binary exists
+		_, err := os.Stat(RtspServerCmd)
+		if err != nil {
+			if os.IsNotExist(err) {
+				return fmt.Errorf("%s file cannot be found: %s", RtspServerCmd, err.Error())
+			}
```
ajcasagrande: what about other error cases? i think this would crash in that scenario

internal/driver/driver.go
```diff
@@ -94,32 +96,52 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
...
+		}
+		rtspProc := exec.Command(RtspServerCmd)
+		rtspProc.Stdout = os.Stdout
+		err = rtspProc.Start()
+		if err != nil {
+			return fmt.Errorf("unable to start %s process: %s", RtspServerCmd, err.Error())
+		}
```
ajcasagrande: I think I would move the starting of the process to the end of this if block instead of the beginning

internal/driver/driver.go
```diff
@@ -94,32 +96,52 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
...
-		rtspServerHostName = DefaultRtspServerHostName
-		d.lc.Warnf("service config %s not found. Use the default value: %s", RtspServerHostName, DefaultRtspServerHostName)
+	var err error
+	d.enableRtspServer, err = strconv.ParseBool(d.ds.DriverConfigs()[EnableRtspServer])
+	if err != nil {
+		return fmt.Errorf("failed to parse EnableRtspServer config: %s", err.Error())
 	}
-	d.lc.Infof("RTSP server hostname: %s", rtspServerHostName)
-	d.rtspHostName = rtspServerHostName
+	if d.enableRtspServer {
```
ajcasagrande: you can think about inverting this to reduce nesting
```go
if !d.enableRtspServer { 
      return nil
}

...
setup rtsp server
```

internal/driver/driver.go
```diff
@@ -140,17 +162,21 @@ func (d *Driver) Start() error {
 	// Make sure the paths of existing devices are up-to-date.
 	go d.RefreshAllDevicePaths()
 
-	d.wg.Add(1)
-	go d.StartRTSPCredentialServer()
+	if d.enableRtspServer {
```
ajcasagrande: Same here, you can invert and return nil, to avoid nesting.

internal/driver/driver.go
```diff
@@ -140,17 +162,21 @@ func (d *Driver) Start() error {
...
+
+		d.wg.Add(1)
+		go d.StartRTSPCredentialServer()
```
ajcasagrande: I know this isn't related to your code, but if you wouldn't mind tweaking this a bit. I prefer the wg.Add and wg.Done to be in close sight of each other. Right now, the wg.Done is in the `StartRTSPCredentialServer` func. I would remove it from there, and do this:

```go
		d.wg.Add(1)
		go func() {
				defer wg.Done()
				d.StartRTSPCredentialServer()
		}
```

this helps to make sure every added waitgroup item is accounted for (and in other situations prevent from having to pass the wg).

ajcasagrande (changes requested):
Looks good! I have a few small suggestions.

internal/driver/driver.go
```diff
@@ -122,6 +133,25 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
...
 
+	// check to see if rtsp-simple-server file/binary exists
+	_, err = os.Stat(RtspServerCmd)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return fmt.Errorf("%s file cannot be found: %s", RtspServerCmd, err.Error())
+		} else if os.IsPermission(err) {
+			return fmt.Errorf("permission denied for %s: %s", RtspServerCmd, err.Error())
+		} else {
+			return fmt.Errorf("failed to validate file %s: %s", RtspServerCmd, err.Error())
```
ajcasagrande: I would change to something like:

`unknown error occurred while checking for rtsp-server binary file %s: %s`

internal/driver/driver.go
```diff
@@ -140,8 +170,16 @@ func (d *Driver) Start() error {
...
+	if !d.enableRtspServer {
+		d.lc.Info("Rtsp server is disabled")
+		return nil
+	}
+
 	d.wg.Add(1)
-	go d.StartRTSPCredentialServer()
+	go func() {
+		defer d.wg.Done()
+		d.StartRTSPCredentialServer()
```
ajcasagrande, line 181: I don't see that you removed the `d.wg.Done()` from `StartRTSPCredentialServer`? If not, please remove it.

ajcasagrande (changes requested):

ajcasagrande (approved):
LGTM

seanohair22 (approved):
LGTM!

cmd/res/configuration.yaml
```diff
@@ -38,6 +38,7 @@ Device:
     Interval: "1h"
 
 Driver:
+  EnableRtspServer: "true"
```
farshidtz, line 41: Is this backwards compatible or will upgrading from v3.0 to v3.1 lead to a disabled RTSP server?

internal/driver/driver.go
```diff
@@ -122,6 +133,25 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
...
+			return fmt.Errorf("unknown error occurred while checking for rtsp-server binary file %s: %s", RtspServerCmd, err.Error())
+		}
+	}
+
+	rtspProc := exec.Command(RtspServerCmd)
+	rtspProc.Stdout = os.Stdout
+	err = rtspProc.Start()
+	if err != nil {
+		return fmt.Errorf("unable to start %s process: %s", RtspServerCmd, err.Error())
+	}
```
farshidtz, line 153: Isn't the process already started in docker-entrypoint: https://github.com/edgexfoundry/device-usb-camera/blob/01aee5a5c41b4c067632ab994dacfc003f161152/docker-entrypoint.sh#L3C1-L4

The snap uses systemctl to start it per app definition here: https://github.com/edgexfoundry/device-usb-camera/blob/01aee5a5c41b4c067632ab994dacfc003f161152/snap/snapcraft.yaml#L35-L39

internal/driver/driver.go
```diff
@@ -122,6 +133,25 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
...
+			return fmt.Errorf("%s file cannot be found: %s", RtspServerCmd, err.Error())
+		} else if os.IsPermission(err) {
+			return fmt.Errorf("permission denied for %s: %s", RtspServerCmd, err.Error())
+		} else {
+			return fmt.Errorf("unknown error occurred while checking for rtsp-server binary file %s: %s", RtspServerCmd, err.Error())
+		}
+	}
+
+	rtspProc := exec.Command(RtspServerCmd)
+	rtspProc.Stdout = os.Stdout
```
farshidtz, line 149: How about standard errors?

farshidtz (commented):
A late review. I'm not sure how this is suppose to work with docker and snaps, in which both have other defined ways of starting the RTSP server.

cmd/res/configuration.yaml
marcpfuller: upgrading to 3.1  would not lead to a disabled rtsp server. 3.0 just will always have rtsp started  by the entrypoint.sh script, while 3.1 just adds the ability to enable or disable rtsp server.

internal/driver/driver.go
marcpfuller: yes, I seem to have forgotten to remove it from the entrypoint script, I will put in a PR to fix this

internal/driver/driver.go
marcpfuller: good point, I will look into having the stderr printed out as well in my next PR

cmd/res/configuration.yaml
farshidtz: I don't think so. The entrypoint.sh of 3.0 will be replaced by that of 3.1 which [should not start RTSP](https://github.com/edgexfoundry/device-usb-camera/pull/268#discussion_r1296004827). When upgrading, the configuration file will not have `EnableRtspServer: "true"` so it will be the zero-value of the boolean instead which is `EnableRtspServer: "false."`

lenny-goodell: This config setting should be a bool not a string, that way it doesn't have to be parsed and has a default value of `false`. Also for backward compatibility it should be `DisableRtspServer` since currently the RTSP Servicer is enabled by default. Please make this change.

farshidtz: Oh, I didn't even notice that it was a string rather than a boolean.

marcpfuller: the SDK only return map[string]string for driver configurations. "

lenny-goodell: > the SDK only return map[string]string for driver configurations. "

Right and the service doesn't have any custom config... So then the code here must consider empty string as `true` to not break backward compatibility rather then return a parse error.
https://github.com/edgexfoundry/device-usb-camera/blob/860b2ed41c63812ea8e78996a55d34ef8e22b126/internal/driver/driver.go#L119-L122

internal/driver/constants.go
```diff
@@ -16,6 +16,8 @@ const (
 	AutoStreaming                   = "AutoStreaming"
 	InputIndex                      = "InputIndex"
 	UrlRawQuery                     = "urlRawQuery"
+	EnableRtspServer                = "EnableRtspServer"
+	RtspServerCmd                   = "./rtsp-simple-server"
```
farshidtz, line 20: This should be made configurable. The executable path can vary depending on the deployment.

See the [snap testing error, line 161](https://github.com/edgexfoundry/device-usb-camera/actions/runs/5861912136/job/15892978209?pr=268#step:2:166):
`
Aug 15 00:10:47 fv-az570-769 edgex-device-usb-camera.device-usb-camera[5200]: level=ERROR ts=2023-08-15T00:10:47.372662735Z app=device-usb-camera source=init.go:77 msg="ProtocolDriver init failed: ./rtsp-simple-server file cannot be found: stat ./rtsp-simple-server: no such file or directory"
`
For the snap, it needs to be `$SNAP/bin/rtsp-simple-server` -> `/snap/edgex-device-usb-camera/current/bin/rtsp-simple-server`

internal/driver/driver.go
farshidtz: Being addressed in https://github.com/edgexfoundry/device-usb-camera/pull/270

internal/driver/driver.go
farshidtz: Being addressed in https://github.com/edgexfoundry/device-usb-camera/pull/270

cmd/res/configuration.yaml
farshidtz: Being addressed in https://github.com/edgexfoundry/device-usb-camera/pull/270

PR #269, opened 2023-08-11
feat: implement get/set pixel format commands.

presatish:

## PR Checklist

  <link to docs PR>

## Testing Instructions

1. Build docker image using `make docker` and deploy all edgex core services along with the usb service.
1. Executing get `PixelFormat` command for a valid camera with a valid path index should give valid result with height, width, pixelFormat and other parameters.
2. Invalid camera or path index (not video streaming path) should give error.
3. Executing set `PixelFormat` command with valid parameters (height, width and pixelFormat) should be successful (200 status code). Postman collection can be referred for example. Execute get command to validate new parameters have been set.
4. Invalid values should give errors.

### DIFF
--- a/cmd/res/profiles/general.usb.camera.yaml
+++ b/cmd/res/profiles/general.usb.camera.yaml
@@ -95,6 +95,14 @@ deviceResources:
     properties:
       valueType: "Object"
       readWrite: "RW"
+  - name: "PixelFormat"
+    description: "Set and get the pixel format, height and width"
+    attributes:
+      getFunction: "VIDEO_GET_PIXELFORMAT"
+      setFunction: "VIDEO_SET_PIXELFORMAT"
+    properties:
+      valueType: "Object"
+      readWrite: "RW"
   - name: "StreamURI"
     description: "Get video-streaming URI."
     attributes:
--- a/cmd/res/profiles/general.usb.camera.yaml
+++ b/cmd/res/profiles/general.usb.camera.yaml
@@ -88,7 +88,7 @@ deviceResources:
       readWrite: "W"
       defaultValue: "false"
   - name: "FrameRate"
-    description: "Set and get the stream frame rate"
+    description: "Get and set the stream frame rate"
     attributes: 
         getFunction: "VIDEO_GET_FRAMERATE"
         setFunction: "VIDEO_SET_FRAMERATE"
@@ -96,7 +96,7 @@ deviceResources:
       valueType: "Object"
       readWrite: "RW"
   - name: "PixelFormat"
-    description: "Set and get the video pixel format"
+    description: "Get and set the video pixel format"
     attributes:
       getFunction: "VIDEO_GET_PIXELFORMAT"
       setFunction: "VIDEO_SET_PIXELFORMAT"
--- a/cmd/res/profiles/general.usb.camera.yaml
+++ b/cmd/res/profiles/general.usb.camera.yaml
@@ -96,7 +96,7 @@ deviceResources:
       valueType: "Object"
       readWrite: "RW"
   - name: "PixelFormat"
-    description: "Set and get the pixel format, height and width"
+    description: "Set and get the video pixel format"
     attributes:
       getFunction: "VIDEO_GET_PIXELFORMAT"
       setFunction: "VIDEO_SET_PIXELFORMAT"
--- a/docs/USB-Camera-Collection.postman_collection.json
+++ b/docs/USB-Camera-Collection.postman_collection.json
@@ -740,7 +740,7 @@
 				"header": [],
 				"body": {
 					"mode": "raw",
-					"raw": "{\n\"PixelFormat\": {\n        \"Width\":\"640\",\n        \"Height\":\"480\",\n        \"PixelFormat\": \"RGB\",\n        \"Field\": \"ANY\",\n        \"BytesPerLine\":\"7\",\n        \"SizeImage\": \"100\",\n        \"Colorspace\":\"DEFAULT\"\n    }\n}",
+					"raw": "{\n\"PixelFormat\": {\n        \"Width\":\"640\",\n        \"Height\":\"480\",\n        \"PixelFormat\": \"YUYV\"\n   }\n}",
 					"options": {
 						"raw": {
 							"language": "json"
--- a/docs/USB-Camera-Collection.postman_collection.json
+++ b/docs/USB-Camera-Collection.postman_collection.json
@@ -51,7 +51,7 @@
 						"{{device_service_name}}"
 					]
 				},
-				"description": "To get all onvif devices change {device_service_name} environment varibale to device-onvif-camera"
+				"description": "To get all USB devices change {device_service_name} environment varibale to device-usb-camera"
 			},
 			"response": []
 		},
--- a/docs/USB-Camera-Collection.postman_collection.json
+++ b/docs/USB-Camera-Collection.postman_collection.json
@@ -666,6 +666,114 @@
 			"response": []
 		},
 		{
+			"name": "Get Pixel Format",
+			"event": [
+				{
+					"listen": "test",
+					"script": {
+						"exec": [
+							"pm.test(\"Status code is 200\", function () {",
+							"  pm.response.to.have.status(200);",
+							"});"
+						],
+						"type": "text/javascript"
+					}
+				}
+			],
+			"protocolProfileBehavior": {
+				"disableBodyPruning": true
+			},
+			"request": {
+				"method": "GET",
+				"header": [],
+				"body": {
+					"mode": "raw",
+					"raw": "",
+					"options": {
+						"raw": {
+							"language": "json"
+						}
+					}
+				},
+				"url": {
+					"raw": "http://{{command_host}}:{{command_port}}/api/v3/device/name/{{camera_name}}/PixelFormat?PathIndex={{path_index}}",
+					"protocol": "http",
+					"host": [
+						"{{command_host}}"
+					],
+					"port": "{{command_port}}",
+					"path": [
+						"api",
+						"v3",
+						"device",
+						"name",
+						"{{camera_name}}",
+						"PixelFormat"
+					],
+					"query": [
+						{
+							"key": "PathIndex",
+							"value": "{{path_index}}"
+						}
+					]
+				}
+			},
+			"response": []
+		},
+		{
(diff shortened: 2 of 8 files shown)


cmd/res/profiles/general.usb.camera.yaml
```diff
@@ -95,6 +95,14 @@ deviceResources:
     properties:
       valueType: "Object"
       readWrite: "RW"
+  - name: "PixelFormat"
+    description: "Set and get the video pixel format"
```
ajcasagrande: For some reason `Get and set` sounds better to me

docs/USB-Camera-Collection.postman_collection.json
```diff
@@ -665,6 +665,114 @@
...
+					"mode": "raw",
+					"raw": "",
+					"options": {
+						"raw": {
+							"language": "json"
+						}
+					}
+				},
+				"url": {
+					"raw": "http://{{command_host}}:{{command_port}}/api/v3/device/name/{{camera_name}}/PixelFormat?PathIndex={{path_index}}",
```
ajcasagrande, line 699: @ChristianDarr-personal do we need to integrate your PR code about the stream filtering with this API?

docs/USB_camera_env.postman_environment.json
```diff
@@ -32,18 +32,6 @@
...
-			"value": "59985",
-			"type": "default",
-			"enabled": true
-		},
-		{
-			"key": "device_onvif_service_host",
-			"value": "localhost",
-			"type": "default",
-			"enabled": true
-		},
```
ajcasagrande, line 46: :+1:

internal/driver/constants.go
```diff
@@ -29,6 +29,9 @@ const (
 	FrameRateValueDenominator       = "FrameRateValueDenominator"
 	FrameRateValueNumerator         = "FrameRateValueNumerator"
 	PathIndex                       = "PathIndex"
+	PixFmtWidth                     = "Width"
+	PixFmtHeight                    = "Height"
+	PixFmtPixFmt                    = "PixelFormat"
```
ajcasagrande: this should probably just be: `PixelFormat = "PixelFormat"` regardless of the two above it

internal/driver/constants.go
```diff
@@ -29,6 +29,9 @@ const (
 	FrameRateValueDenominator       = "FrameRateValueDenominator"
 	FrameRateValueNumerator         = "FrameRateValueNumerator"
 	PathIndex                       = "PathIndex"
+	PixFmtWidth                     = "Width"
+	PixFmtHeight                    = "Height"
```
ajcasagrande: Do these need to be prefixed, or can they just be `Width = "Width"` and `Height = "Height"`?

internal/driver/driver.go
```diff
@@ -317,6 +317,12 @@ func (d *Driver) ExecuteReadCommands(device *Device, req sdkModels.CommandReques
 			return cv, errorWrapper.CommandError(command, err)
 		}
 		cv, err = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeObject, data)
+	case VideoGetPixelFormat:
+		data, err = device.GetPixelFormat(cameraDevice)
+		if err != nil {
+			return cv, errorWrapper.CommandError(command, err)
```
```suggestion
			return nil, errorWrapper.CommandError(command, err)
```

internal/driver/device.go
```diff
@@ -138,6 +193,44 @@ func (dev *Device) GetFrameRate(usbDevice *usbdevice.Device) (v4l2.Fract, error)
...
+	result.Field = v4l2.Fields[pixFmt.Field]
+	result.BytesPerLine = pixFmt.BytesPerLine
+	result.SizeImage = pixFmt.SizeImage
+	result.Colorspace = v4l2.Colorspaces[pixFmt.Colorspace]
+	result.Priv = pixFmt.Priv
+	result.Flags = pixFmt.Flags
+	result.YcbcrEnc = v4l2.YCbCrEncodings[pixFmt.YcbcrEnc]
+	result.HSVEnc = v4l2.YCbCrEncodings[pixFmt.HSVEnc]
+	result.Quantization = v4l2.Quantizations[pixFmt.Quantization]
+	result.XferFunc = v4l2.XferFunctions[pixFmt.XferFunc]
```
ajcasagrande: this can use struct initializer:

```go
result := PixelFormat{
	Height: pixFmt.Height,
	Width: pixFmt.Width,
	...
}   
```

ajcasagrande (changes requested):

internal/driver/types.go
```diff
@@ -31,3 +31,36 @@ type FrameRateFormat struct {
...
+	"JPEG":  v4l2.PixelFmtJPEG,
+	"MPEG":  v4l2.PixelFmtMPEG,
+	"H264":  v4l2.PixelFmtH264,
+	"MPEG4": v4l2.PixelFmtMPEG4,
+	"UYVY":  v4l2.PixelFmtUYVY,
+	// pixel formats not directly included as part of v4l2 pixel format definitions
+	"BYR2": 844257602,
+	"Z16":  540422490,
+	"Y8I":  541669465,
+	"Y12I": 1228026201,
```
ajcasagrande: You should have these defined as constants which are referenced here. May need to coordinate with @ChristianDarr-personal's PR

internal/driver/constants.go
presatish: This leads to duplicate variable but I will change to something better.

internal/driver/driver.go
presatish: But other commands in the same file are returning `cv` only which should be fine as it will be nil in case of error?

internal/driver/types.go
presatish: I looked into his , ya I will change into constants. Not sure why he has even descriptions as constants as we can retrieve them using some v4l2 lib funcs.

internal/driver/driver.go
ajcasagrande: In Marc's PR he is changing them all to `nil` so this will be consistent with that when merged

internal/driver/constants.go
```diff
@@ -81,4 +81,10 @@ const (
 	UdevSerialShort = "ID_SERIAL_SHORT"
 	UdevSerial      = "ID_SERIAL"
 	UdevV4lProduct  = "ID_V4L_PRODUCT"
+
+	// Pixel Formats not supported by go4vl pre-defined pixel format definitions
+	BYR2PixFmt     = 844257602
+	DepthZ16PixFmt = 540422490
+	Y8IPixFmt      = 541669465
+	Y12IPixFmt     = 28026201
```
ajcasagrande: I would put the `PixFmt` as a prefix instead of suffix. `PixFmtBYR2`.. etc

internal/driver/constants.go
```diff
@@ -29,9 +29,9 @@ const (
 	FrameRateValueDenominator       = "FrameRateValueDenominator"
 	FrameRateValueNumerator         = "FrameRateValueNumerator"
 	PathIndex                       = "PathIndex"
-	PixFmtWidth                     = "Width"
-	PixFmtHeight                    = "Height"
-	PixFmtPixFmt                    = "PixelFormat"
+	Width                           = "Width"
+	Height                          = "Height"
+	PixFormat                       = "PixelFormat"
```
ajcasagrande: I think I would recommend just removing the PixFormat const and use the existing PixelFormat one as otherwise the name and values do not match.

ajcasagrande (changes requested):

docs/USB-Camera-Collection.postman_collection.json
```diff
@@ -665,6 +665,114 @@
...
+					"path": [
+						"api",
+						"v3",
+						"device",
+						"name",
+						"{{camera_name}}",
+						"PixelFormat"
+					],
+					"query": [
+						{
```
ChristianDarr-personal, line 766: It might be best to go ahead and add the StreamFormat key value pair here

ajcasagrande (dismissed):

docs/USB-Camera-Collection.postman_collection.json
ChristianDarr-personal: Yes we should, it'll probably be easiest to do it now

internal/driver/device.go
```diff
@@ -91,6 +92,60 @@ func (dev *Device) StopStreaming() {
...
+		height, err := strconv.ParseUint(heightValue.(string), 0, 32)
+		if err != nil {
+			return fmt.Errorf("invalid input: error parsing height for the device %s, error: %s", dev.name, err)
+		}
+		v4l2PixelFormat.Height = uint32(height)
+	}
+
+	pixFormatValue, ok := params.(map[string]interface{})[PixelFormat]
+	if ok {
+		pixelFormat, ok := PixelFormatV4l2Mappings[fmt.Sprint(pixFormatValue)]
```
ChristianDarr-personal: Why do you use fmt.sprint instead of .(string) here?

docs/USB-Camera-Collection.postman_collection.json
presatish: @ChristianDarr-personal But for that I need to get all the related code which is still in review. I think it would be better to incorporate this after you merge?

internal/driver/device.go
presatish: I think Mudit added this but this is not required even though its the same result. I will change it.

docs/USB-Camera-Collection.postman_collection.json
presatish: You mean for now to just add in the doc i.e. here?

ChristianDarr-personal: Yeah it would be easiest to just include the query parameters in your new apis so we don't have to create a new pr after ours merge. Check my pr for the formatting

internal/driver/device.go
ChristianDarr-personal: It's not a big deal, it was just inconsistent with the others.

docs/USB-Camera-Collection.postman_collection.json
ChristianDarr-personal: I don't think there is any of my code needed, if they merge at the same time, the only thing we need to adapt is the postman collection

ChristianDarr-personal (approved):

PR #270, opened 2023-08-22
fix: support external RTSP servers using RtspServerMode

marcpfuller:

## PR Checklist

- [ ] I have added unit tests for the new feature or bug fix (if not, why?) n/a
https://github.com/edgexfoundry/edgex-docs/pull/1203

## Testing Instructions
1. run `make test`
2. verify all tests pass
3. start up edgex via edgex-compose with no-secty: `make run no-secty`
4. run device-usb-camera with `disableRtspServer` set to false `./device-usb-camera -cp -d -o`
5. verify RTSP server starts
6. run device-usb-camera with `disableRtspServer` set to true `./device-usb-camera -cp -d -o`
7. verify RTSP server does not start


marcpfuller:
This PR addresses comments made after merge for #268

lenny-goodell (dismissed):
LGTM

farshidtz (dismissed):
Looks good.

Note that the snap tests are still failing because the RTCP binary path has been hardcoded. See https://github.com/edgexfoundry/device-usb-camera/pull/268#discussion_r1302646697

internal/driver/driver.go
```diff
@@ -134,22 +141,24 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
 	}
 
 	// check to see if rtsp-simple-server file/binary exists
-	_, err = os.Stat(RtspServerCmd)
+	rtspExecutable := d.ds.DriverConfigs()[RtspServerExe]
```
lenny-goodell, line 167: Should handle not set in a better way, like check for empty and return error that it isn't set

lenny-goodell (dismissed):

marcpfuller:
@farshidtz  I have made the file path to rtsp server  a configurable. Snap no longer needs to start rtsp server, and you will need to do a environment variable override for the snap location of rtsp simple server. The environment variable override would be `DRIVER_RTSPSERVEREXECUTABLE`

internal/driver/driver.go
```diff
@@ -97,11 +97,18 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
 	}
 
 	var err error
-	d.enableRtspServer, err = strconv.ParseBool(d.ds.DriverConfigs()[EnableRtspServer])
-	if err != nil {
-		return fmt.Errorf("failed to parse EnableRtspServer config: %s", err.Error())
+	// if DisableRtspServer config id blank disableRtspServer should be false
```
presatish:
```suggestion
	// if DisableRtspServer config parameter is empty then it should be set to false
```

internal/driver/driver.go
```diff
@@ -97,11 +97,18 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
...
+	disableRtspServer := d.ds.DriverConfigs()[DisableRtspServer]
+	if len(disableRtspServer) == 0 {
+		d.disableRtspServer = false
+	} else {
+		d.disableRtspServer, err = strconv.ParseBool(d.ds.DriverConfigs()[DisableRtspServer])
+		if err != nil {
+			return fmt.Errorf("failed to parse EnableRtspServer config: %s", err.Error())
```
presatish:
```suggestion
			return fmt.Errorf("failed to parse DisableRtspServer config: %s", err.Error())
```

internal/driver/driver.go
```diff
@@ -97,11 +97,18 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
 	}
 
 	var err error
-	d.enableRtspServer, err = strconv.ParseBool(d.ds.DriverConfigs()[EnableRtspServer])
-	if err != nil {
-		return fmt.Errorf("failed to parse EnableRtspServer config: %s", err.Error())
+	// if DisableRtspServer config parameter is empty then it should be set to false
+	disableRtspServer := d.ds.DriverConfigs()[DisableRtspServer]
+	if len(disableRtspServer) == 0 {
```
```suggestion
	if disableRtspServer == "" {
```

internal/driver/driver.go
```diff
@@ -97,11 +97,18 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
 	}
 
 	var err error
-	d.enableRtspServer, err = strconv.ParseBool(d.ds.DriverConfigs()[EnableRtspServer])
-	if err != nil {
-		return fmt.Errorf("failed to parse EnableRtspServer config: %s", err.Error())
+	// if DisableRtspServer config parameter is empty then it should be set to false
```
```suggestion
	// If DisableRtspServer config parameter is empty, then it should be
	// set to false to retain backwards compatibility
```

internal/driver/driver.go
```diff
@@ -134,22 +141,28 @@ func (d *Driver) Initialize(sdk interfaces.DeviceServiceSDK) error {
 	}
 
 	// check to see if rtsp-simple-server file/binary exists
-	_, err = os.Stat(RtspServerCmd)
+	rtspExecutable := d.ds.DriverConfigs()[RtspServerExe]
+	if len(rtspExecutable) == 0 {
```
```suggestion
	if rtspExecutable == "" {
```

ajcasagrande (changes requested):
Unfortunately there are still some issues regarding the architectural design behind this story.

When `DisableRtspServer` is set to true, then all REST APIs related to streaming will return errors (as designed). However, this does not account for the case where the user wants to disable the internal rtsp server because they want to run it separately (via docker, snap, on a different machine, etc).

In those cases, the user would (optionally) set the `RtspServerHostName` to something different, and disable the internal rtsp server.  I am not sure the best way to handle re-enabling the REST APIs.

Some options:
- Add another config var to specify that an external one is in use
- Change `DisableRtspServer` to be an enum `RtspServerMode` with:
  - `internal` (or `enabled`) - run server and enable apis
  - `external` - do not run server, but enable apis
  - `disabled` - do not run server and disable apis

Open to recommendations @farshidtz @lenny-intel

farshidtz:
> Unfortunately there are still some issues regarding the architectural design behind this story.
> 
> When `DisableRtspServer` is set to true, then all REST APIs related to streaming will return errors (as designed). However, this does not account for the case where the user wants to disable the internal rtsp server because they want to run it separately (via docker, snap, on a different machine, etc).
> 
> In those cases, the user would (optionally) set the `RtspServerHostName` to something different, and disable the internal rtsp server. I am not sure the best way to handle re-enabling the REST APIs.
> 
> Some options:
> 
> * Add another config var to specify that an external one is in use
> * Change `DisableRtspServer` to be an enum `RtspServerMode` with:
>   
>   * `internal` (or `enabled`) - run server and enable apis
>   * `external` - do not run server, but enable apis
>   * `disabled` - do not run server and disable apis
> 
> Open to recommendations @farshidtz @lenny-intel

Is there a use case where users would use this device service without RTSP server and the streaming APIs? 
If not, then the enum could be `RtspServer: internal|external`.

lenny-goodell:
> Some options:
> 
> * Add another config var to specify that an external one is in use
> * Change `DisableRtspServer` to be an enum `RtspServerMode` with:
>   
>   * `internal` (or `enabled`) - run server and enable apis
>   * `external` - do not run server, but enable apis
>   * `disabled` - do not run server and disable apis

I like RtspServerMode with values:

- `internal` - run server and enable apis
- `external` - do not run server, but enable apis
- `none` - do not run server and disable apis
- `` - Defaults to `internal` 

> Is there a use case where users would use this device service without RTSP server and the streaming APIs?
@farshidtz , yes when only need control features and/or accessing the /dev device directly.

ajcasagrande:
> I like RtspServerMode with values:
> 
> * `internal` - run server and enable apis
> * `external` - do not run server, but enable apis
> * `none` - do not run server and disable apis
> * `` - Defaults to `internal`

I like this approach, and if the user specifies anything out of this range, we can error out as bad configuration

ajcasagrande:
Ready for review again

seanohair22 (approved):
LGTM! Tested and works with valid values and catches bad config values!

presatish (approved):
LGTM

ajcasagrande (approved):

PR #271, opened 2023-08-23
feat: Add proxy to build script for device-usb-service

seanohair22:

## PR Checklist

[  Docs PR for proxy](https://github.com/edgexfoundry/edgex-docs/pull/1184)

## Testing Instructions

Add proxy to docker (ping me for specific proxy setup instructions)
Test onsite or on VPN
Run `make docker`, the service should build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,9 @@ docker:
 	docker build . \
 		--build-arg ADD_BUILD_TAGS=$(ADD_BUILD_TAGS) \
 		--label "git_sha=$(GIT_SHA)" \
+		--build-arg http_proxy=$(HTTP_PROXY) \
+		--build-arg https_proxy=$(HTTPS_PROXY) \
+		--build-arg no_proxy=$(NO_PROXY) \
 		-t edgexfoundry/device-usb-camera:$(GIT_SHA) \
 		-t edgexfoundry/device-usb-camera:$(VERSION)-dev
 


Makefile
```diff
@@ -33,6 +33,9 @@ docker:
 	docker build . \
 		--build-arg ADD_BUILD_TAGS=$(ADD_BUILD_TAGS) \
 		--label "git_sha=$(GIT_SHA)" \
+		--build-arg http_proxy=$(HTTP_PROXY) \
+		--build-arg https_proxy=$(HTTPS_PROXY) \
+		--build-arg no_proxy=$(NO_PROXY) \
```
ajcasagrande, line 38: Should these use the lowercase form as well? aka `--build-arg http_proxy=$(http_proxy)`, which i think can just be shortened to `--build-arg http_proxy`

presatish (commented):
The changes look fine but I am not able to test as it looks like vpn certificate for my linux machine has been expired or has some issues because of which I am unable to test.

Makefile
seanohair22: I followed what was present in the device-onvif-service, that way the environment variables only need to be set with one format for building both services...

ajcasagrande: ok, that seems fair, as long as the documentation matches up, which I believe it does when I reviewed that last week.

ajcasagrande (approved):

PR #290, opened 2023-09-15
fix: Both path and paths need to be supported to address the breaking change until the next release

### LINKED ISSUE #278: Config variable Path changed to Paths to support real sense cameras
# 🐞 Bug Report

device-usb-camera

### Is this a regression?
No

After starting all the core services and the usb service when the usb device is manually added using `Path` instead of `Paths` this issue is seen.

## 🔥 Exception or Error
<pre><code>
curl -X POST -H 'Content-Type: application/json'  \
http://localhost:59881/api/v3/device \
-d '[
    {
    "apiVersion" : "v3",
    "device": {
        "name": "Camera001",
        "serviceName": "device-usb-camera",
        "profileName": "USB-Camera-General",
        "description": "My test camera",
        "adminState": "UNLOCKED",
        "operatingState": "UP",
        "protocols": {
            "USB": {
            "CardName": "UVC Camera (012d:0678)",
            "SerialNumber": "71C0AE6","
            "Path": "/dev/video0",
            "AutoStreaming": "false"
            }
        }
    }
    }
]'

[{"apiVersion":"v3","message":"Device Camera001 validation failed: property Paths of protocol USB is missing. Please check device configuration","statusCode":500}]

</code></pre>

## 🌍  Your Environment
**Deployment Environment:** Docker, Docker + Dev Hybrid

**EdgeX Version [**REQUIRED**]:** main

**Anything else relevant?**
`Path` `Deprecated` info needs to be added in the code and docs. For now the code has to support both `Path` and `Paths` to avoid breaking change until the next release.

presatish:

## PR Checklist

  <link to docs PR>

## Testing Instructions
- make test
- make build
- make docker
- Get the edgex core and device usb service running in 3.0 version (Minnesota)
- The usb devices get registered in edgex with `Path` as one of the protocol properties which is a string.
- Stop the containers without deleting the volume (make down)
- Checkout the main version of edgex compose and get the core services running.
-  Get this version of device service running natively or even in docker
-  The same devices get discovered again and `Path` changes to `Paths` which is an array of string.


internal/driver/helper.go
```diff
@@ -14,23 +14,17 @@ import (
...
+var userPassRegex = regexp.MustCompile(`//(\S+):(\S+)@`)
 
 type EdgeXErrorWrapper struct{}
 
 func (e EdgeXErrorWrapper) CommandError(command string, err error) errors.EdgeX {
 	return errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf("failed to execute %s command", command), err)
 }
 
-// redact removes all instances of basic auth (ie. rtsp://username:password@server) from a url
+// redact removes all instances of basic auth (i.e. rtsp://username:password@server) from an url
```
ajcasagrande: `a url` is the correct grammer here: https://www.techtarget.com/whatis/feature/Which-is-correct-a-URL-or-an-URL

internal/driver/helper.go
```diff
@@ -14,23 +14,17 @@ import (
 	"github.com/edgexfoundry/go-mod-core-contracts/v3/errors"
 )
 
-const (
-	redactedStr = "//<redacted>@"
```
ajcasagrande, line 18: I feel this makes more sense being defined here, closer to its usage

internal/driver/types.go
```diff
@@ -10,12 +10,6 @@ import "github.com/vladimirvivien/go4vl/v4l2"
 
 type RTSPServerMode string
 
-const (
-	RTSPServerModeInternal RTSPServerMode = "internal"
-	RTSPServerModeExternal RTSPServerMode = "external"
-	RTSPServerModeNone     RTSPServerMode = "none"
-)
```
ajcasagrande: these belong to the `type RTSPServerMode string` above, and implement an enum.

internal/driver/driver.go
```diff
@@ -727,10 +703,33 @@ func (d *Driver) RefreshAllDevicePaths() {
...
+			}
+		}
+	}
+}
+
 // RefreshDevicePaths checks whether the existing device matches the connected device.
 // If there is a mismatch between them, scan all paths to find the matching device
 // and update the existing device with the correct path.
 func (d *Driver) RefreshDevicePaths(cd models.Device) {
+	d.updatePathToPaths(cd)
```
ajcasagrande, line 731: Is this the only place that needs to call this?

internal/driver/driver.go
```diff
@@ -727,10 +703,33 @@ func (d *Driver) RefreshAllDevicePaths() {
...
+// Deprecated: 3.0 version of the service supports Path which is a string instead of Paths which is []string which leads to breaking change.
+// updatePathToPaths takes care of this breaking change.
+func (d *Driver) updatePathToPaths(cd models.Device) {
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+
+	if value, ok := cd.Protocols[UsbProtocol][Path]; ok {
+		if value != nil {
+			cd.Protocols[UsbProtocol][Paths] = []string{value.(string)}
+			delete(cd.Protocols[UsbProtocol], "Path")
```
ajcasagrande, line 715: I thought we had discussed keeping both fields in the metadata because otherwise, when you get a device it will be missing the `Path` field that used to be there. @lenny-intel thoughts?

ajcasagrande (changes requested):

internal/driver/helper.go
presatish: Ok, I thought it would be better to keep all constants in one place.

internal/driver/driver.go
presatish: yes this function is called when the service starts up and also when the discovery is called.

internal/driver/driver.go
presatish: I don't remember discussing about this, you mean creating `Metadata` section within protocols like how we have in onvif service? Existing `Path` field will be changed to `Paths` with this code which is easier to maintain. We have added in the docs too that `Path` field is deprecated and users should be using `Paths`. I think @lenny-intel was fine with this.

internal/driver/helper.go
presatish: Even the silly IDE suggested to me that it should be `an` :-)

internal/driver/driver.go
ajcasagrande: I mean keeping bath `Path` and `Paths` in the `Onvif` protocol properties, and using code to sync them. As long as @lenny-intel is OK with this approach I am fine as well, since this will be cleaner.

internal/driver/types.go
```diff
@@ -73,3 +80,25 @@ var PixelFormatV4l2Mappings = map[string]uint32{
...
+	v4l2.PixelFmtH264:  RGB,
+	v4l2.PixelFmtMPEG4: RGB,
+	v4l2.PixelFmtUYVY:  Greyscale,
+	PixFmtBYR2:         RGB,
+	PixFmtY8I:          Greyscale,
+	PixFmtY12I:         Greyscale,
+	PixFmtDepthZ16:     Depth,
+}
+
+const (
```
ajcasagrande: I would put these back where they were, or move the `type` definition down here. I like them being together.

ajcasagrande (changes requested):

internal/driver/driver.go
lenny-goodell: Yes, we have to keep both. My thought is if a device is loaded that has Path set, then set Paths to the single Path. That way the driver code always deals with Paths.

ajcasagrande (approved):
LGTM

vyshali-chitikeshi:
Though its already merged into main, I did validaiton to ensure end-to-end testing (discovering usb cameras, adding them egex metadata and rest-api's) works correctly both with 'path' and 'paths' fields and it looks good. 
Validated with below combinations of code:
1. Latest (main branch) egex and Latest (main branch) USB code - works with 'paths'
2. Latest (main branch) egex and 3.0  USB code - works with 'path'
3. 3.0 egex and Latest (main branch) USB code - works with 'paths'
4.  3.0 egex and 3.0 USB code - works with 'path

attaced scrresnhots for refecne.
[latest_edgex_and_3.0_usb_device_usb.txt](https://github.com/edgexfoundry/device-usb-camera/files/12742218/latest_edgex_and_3.0_usb_device_usb.txt)
[3.0_edgex_and_latest_usb_device_path.txt](https://github.com/edgexfoundry/device-usb-camera/files/12742219/3.0_edgex_and_latest_usb_device_path.txt)
[3.0_edgex_and_3.0_usb_device_path.txt](https://github.com/edgexfoundry/device-usb-camera/files/12742220/3.0_edgex_and_3.0_usb_device_path.txt)
[latest_edgex_and_latest_usb_device_path.txt](https://github.com/edgexfoundry/device-usb-camera/files/12742221/latest_edgex_and_latest_usb_device_path.txt)

